PR #10, opened 2016-01-17
Mark stack as non-executable

sebastinas:
We have received a bug report mentioning that the shared library is built with an executable stack (see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=793320 for more details). This PR contains a patch by Russell Coker to remove the executable bit.

### DIFF
--- a/src/pack_x64.s
+++ b/src/pack_x64.s
@@ -6,6 +6,7 @@
 ##      Distributed under the BSD Software License (see license.txt)      ##
 ############################################################################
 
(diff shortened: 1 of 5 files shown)


dbry:
Thanks for the patch! I am currently working in the "unicode" branch, and so just to be clean I'll wait until after I merge back to Master to pull this (which should not be long).

dbry:
Merged. Thanks for the patch!

PR #17, opened 2017-01-20
enable travis ci build and testing

stephengroat:
if you're interested, you could enable travis ci on this repo (https://travis-ci.org/stephengroat/WavPack). I know that i currently only have a very small subset of tests running, but I'm not 100% sure how to get the granularity I need for travis (maximum # of jobs is 200, maximum runtime on each job is 50 mins). More tests could be specific by creating more versions of the `WVTEST_ARGS` environment variable

If you're interested, let me know, i'd be interested in helping.

Also could enable automated Windows builds with [appveyor](https://www.appveyor.com/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: c
 sudo: false
 script:
-  - ./autogen.sh --enable-tests && make && .libs/wvtest --default
+  - ./autogen.sh --enable-tests && make && cli/wvtest --default
 compiler:
   - clang
   - gcc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,13 @@
 language: c
 sudo: false
 script:
-  - ./autogen.sh --enable-tests && make && cli/wvtest --default
+  - ./autogen.sh --enable-tests && make && cli/wvtest $WVTEST_ARGS
 compiler:
   - clang
   - gcc
 os:
   - linux
   - osx
+env:
+  matrix:
+    - WVTEST_ARGS=--default --no-hybrid --no-lossy
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: c
+sudo: false
+script:
+  - ./autogen.sh --enable-tests && make && .libs/wvtest --default
+compiler:
+  - clang
+  - gcc
+os:
+  - linux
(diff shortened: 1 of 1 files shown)


dbry:
Hey, this is pretty cool! I had not heard of Travis up to now, but I looked around and got some idea what's going on. Thanks for introducing me to this.

So, I would just pull in your file, sign into Travis and point it to my repo?

I think a better combination of options for wvtest would be “wvtest --exhaustive –short –no-extras”. That's cuts the time almost in half compared to what you've got set there, and actually covers way more stuff (148 vs 48 tests).

One thing I read is that you can set the Ubuntu dist to “trusty” and that includes Clang 3.5, which works for WavPack (unlike, as I guess you discovered, 3.4). Obviously it would be nice to get both compilers covered.

One thing that makes me a little hesitant is that WavPack development is basically all done by a single person (me) and is also pretty much finished at this point (feature complete), so I'm not sure how useful having continuous integration would be. Of course, if there's no downside hassle and it's free, I guess there's no harm, right?

Thanks again!

stephengroat:
I moved to `trusty` for `clang`, seems like it's working

If you pull the file in, then all you have to do is setup an account with Travis and enable it for the repo (https://docs.travis-ci.com/user/for-beginners). This example shows php, but if you have the `.travis.yml` in the repo, this one is setup for `language: c`

I think that 148 tests is going to hit the travis timeout (60 minutes). If there any way to specify less tests? the `build: matrix` fields allow you to create multiple travis jobs, each with 60 minute time limits. with about 60 or 80 tests in a job, it would be possible to do the complete exhaustive testing. I'm just not 100% sure what other configuration parameters I need to include to split the tests up into 60-80 test units. make sense?

I just like travis because it shows people how to build the software (which is nice). I also think that people who distribute binary distributions of your stuff (in particular, homebrew) really should test it before distributing. travis shows them how to (just my rant, i recently got burned by an untested binary)

dbry:
Okay, cool, I'll try to being this in and test it out.

What I was saying about the tests is that the 148 tests I mentioned should go much faster than the 48 you were doing (because of the --no-extras option). But I'll try it and see.

stephengroat:
Great! seems like those work well!

PR #18, opened 2017-01-23
Appveyor

stephengroat:
It's like travis for windows, you can see builds [here](https://ci.appveyor.com/project/stephengroat/wavpack)


dbry:
Thanks, this is also cool, but I'll have to wait until I update my Windows build environment to something more recent (or try to use the VS 2008 in appveyor). I certainly don't want to be maintaining two sets of project files.

The other issue is that I do most of my development in Linux, and usually during that time the Windows build is broken (which would be annoying if I was reminded of that every push). Once I finish I bring everything into Windows, so it really makes a lot of sense to have the testing all going on on the LInux side.

PR #20, opened 2017-02-18
configure: Restrict arm assembly to armv7 only.

aballier:
ARM assembly in wavpack is armv7 only it seems.
I have reports this causes build failures on armv5: https://bugs.gentoo.org/show_bug.cgi?id=609168


dbry:
I think that the **ssat** instruction should work with armv6 and up, but I'll pull in this fix and verify that later. Thanks!!

dbry:
Had trouble with the merge doing it right from here, but it seems to have gone in this time...

dbry:
I have pushed a better fix for this. I replaced the SSAT with a pair of shifts, so this should now work on all architectures, and so I went back to arm* in the configure. Thanks again!

PR #22, opened 2017-10-12
Fixed compilation errors when using a C++ compiler.

jrlanglois:
## What have I done?

In order to compile the C code in a C++ project, I've had to C-style cast anything that was of type `void*` to the destination type.

The issue we were running into is the blasted [error C2440](https://msdn.microsoft.com/query/dev14.query?appId=Dev14IDEF1&l=EN-US&k=k(C2440)&rd=true).

## Why have I done this? What was I trying to do?

A bit of background: I'm a C++ developer using [JUCE](https://github.com/WeAreROLI/JUCE). In-house at [AudioWorks](https://github.com/Audioworks) we have our own [Unity Build modules](http://buffered.io/posts/the-magic-of-unity-builds/), following similar conventions to JUCE's modules.

We are trying to use this repository as a submodule in order to avoid the whole _Not Invented Here_ problem, and having a facilitated way of keeping up to date. To do this, we have wrapped WavPack's source in a derived implementation of [`juce::AudioFormatReader`](https://www.juce.com/doc/classAudioFormatReader) using a similar approach to JUCE's [OGG](https://github.com/WeAreROLI/JUCE/blob/master/modules/juce_audio_formats/codecs/juce_OggVorbisAudioFormat.cpp#L52-L78) and [FLAC ](https://github.com/WeAreROLI/JUCE/blob/master/modules/juce_audio_formats/codecs/juce_FlacAudioFormat.cpp#L137-L152) support: including the `.c` files directly.
***

If my pull request is completely missing the mark, or if I've gone about this the wrong way (eg: not quite 'getting' C, coding style, wrong repo), let me know what I need to change - assuming you're interested.

...and if the motivation is too left-field, feel free to close the pull request.

Cheers!

### DIFF
--- a/src/decorr_utils.c
+++ b/src/decorr_utils.c
@@ -32,7 +32,7 @@
 int read_decorr_terms (WavpackStream *wps, WavpackMetadata *wpmd)
 {
     int termcnt = wpmd->byte_length;
-    unsigned char *byteptr = wpmd->data;
+    unsigned char *byteptr = (unsigned char *)wpmd->data;
     struct decorr_pass *dpp;
 
     if (termcnt > MAX_NTERMS)
@@ -61,7 +61,7 @@ int read_decorr_terms (WavpackStream *wps, WavpackMetadata *wpmd)
 int read_decorr_weights (WavpackStream *wps, WavpackMetadata *wpmd)
 {
     int termcnt = wpmd->byte_length, tcount;
-    char *byteptr = wpmd->data;
+    char *byteptr = (char *)wpmd->data;
     struct decorr_pass *dpp;
 
     if (!(wps->wphdr.flags & MONO_DATA))
@@ -93,7 +93,7 @@ int read_decorr_weights (WavpackStream *wps, WavpackMetadata *wpmd)
 
 int read_decorr_samples (WavpackStream *wps, WavpackMetadata *wpmd)
 {
-    unsigned char *byteptr = wpmd->data;
+    unsigned char *byteptr = (unsigned char *)wpmd->data;
     unsigned char *endptr = byteptr + wpmd->byte_length;
     struct decorr_pass *dpp;
     int tcount;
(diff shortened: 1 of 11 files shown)


dbry:
I'm not totally against this patch, because it obviously doesn't break anything, but I am a little hesitant. I am very familiar with the C++ requirement for these casts, but all the build systems I've worked on have always automatically compiled .c files as C language, and it's certainly a little weird to require that C code adhere to C++ rules.

On Visual Studio, for example, you can go into `Configure Properties → C/C++ → Advanced → Compile as` to specify C compilation even for files named .cpp.

I also just tried `-fpermissive` on gcc and it converted this error into a warning, which might work for you. This also works inline as a pragma: `#pragma GCC diagnostic warning "-fpermissive"` 

Does this requirement have something to do with the Unity build? Are the C and C++ files all smushed together?

Another issue is that even if I incorporated this change, I would probably break the code when I make changes in the future because none of my build configurations check for this error, and I'm sure I'd forget to include this cast at some point (especially because I believe that the whole point of using a void pointer is to avoid silly casts like these).

However, like I said, I'm not totally against this and obviously I'd like to accommodate anyone who is adding WavPack support to an application! If you can't figure out another way around this, I'll go ahead and incorporate the patch.

Thanks!

David

jrlanglois:
> Does this requirement have something to do with the Unity build? Are the C and C++ files all smushed together?

Yep, that's generally it.

> Another issue is that even if I incorporated this change, I would probably break the code when I make changes in the future because none of my build configurations check for this error, and I'm sure I'd forget to include this cast at some point (especially because I believe that the whole point of using a void pointer is to avoid silly casts like these).

I've no issues with helping out in this department! Seeing that we're including the files directly from a submodule of this repository, an update to the submodule would mean fixing any errors/warnings that come up.

dbry:
Okay, cool!

PR #23, opened 2017-10-20
Updated README to Markdown format.

jrlanglois:
The goal of this change is to lower the barrier to entry.

By modernising the formatting a bit, and providing direct links to various pieces of documentation and referenced files, it should be much easier to get started and to keep track of the current state of things.

To see the formatting in action, [view the README.md file here](https://github.com/jrlanglois/WavPack/blob/b572d7f36dc8ca5cc109d35b2237fa620d982096/README.md).


dbry:
Very cool...thanks for making this!

The only issue is that it doesn't seem to be copied into the distribution (make dist-bzip2), and I can't find where README is hardcoded. If I can't figure that out I may need to include both versions.

jrlanglois:
Awesome, thanks for merging this in!

PR #34, opened 2018-06-06
General Sprucing Up

jrlanglois:
Just a quick pass at summer (?) cleaning.

* Added missing active branch to README.md.
* Trimmed all whitespace in code files.
* Updated copyrights.

Hopefully it's not too intrusive!


dbry:
Thanks for doing this! I'll be happy to merge it, but I think that the whitespace fixes should _only_ be applied to _my_ files and not files I obtained (like all those in winamp/wasabi, for example) and it's probably not great to manually edit .rc files. I looked briefly for some way to selectively merge only the files I wanted but couldn't find it (I'm not GitHub guru, that's for sure). Want to create a pared-down version? If not, I can do it manually by just grabbing the edits I want.

dbry:
Thanks again...I just pulled this in manually so I could cherry-pick the files I wanted.

PR #39, opened 2018-07-12
fix typo

yangfl:


dbry:
Thanks for this! I didn't think it was right to bring in the wasabi changes (because they aren't my files) so I just brought in the others by hand. Cheers!

PR #46, opened 2018-09-04
Fix coding defects

Chocobo1:
Hi!
I scanned the code base with cppcheck and it found some defects.
Please take a look if I fixed them right.

### DIFF
--- a/src/open_raw.c
+++ b/src/open_raw.c
@@ -156,7 +156,7 @@ WavpackContext *WavpackOpenRawDecoder (
         }
 
         while (main_bytes >= 12) {
-            WavpackHeader *wphdr = malloc (sizeof (WavpackHeader));
+            WavpackHeader *wphdr;
 
             if (!msi) {
                 block_samples = *mcp++;
@@ -201,6 +201,7 @@ WavpackContext *WavpackOpenRawDecoder (
                 return NULL;
             }
 
+            wphdr = malloc (sizeof (WavpackHeader));
             memset (wphdr, 0, sizeof (WavpackHeader));
             memcpy (wphdr->ckID, "wvpk", 4);
             wphdr->ckSize = sizeof (WavpackHeader) - 8 + block_size;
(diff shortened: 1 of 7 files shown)


src/open_raw.c
```diff
@@ -201,6 +199,7 @@ WavpackContext *WavpackOpenRawDecoder (
                 return NULL;
             }
 
+            WavpackHeader *wphdr = malloc (sizeof (WavpackHeader));
             memset (wphdr, 0, sizeof (WavpackHeader));
```
dbry: Putting declarations in the middle of a block is C99, and not universally portable. Specifically, it doesn't work with Visual Studio 2008, which I still use.

winamp/in_wv.c
```diff
@@ -1868,7 +1868,7 @@ int WavPack_GetAlbumArt(const wchar_t *filename, const wchar_t *type, void **bit
 			memmove (buffer, buffer + i + 1, *len);
 			retval = 0;
 #ifdef DEBUG_CONSOLE
-			sprintf (error, "WavPack_GetAlbumArt (\"%s\", %d) success!\n", wide2char (*mime_type), *len);
+			sprintf (error, "WavPack_GetAlbumArt (\"%s\", %zu) success!\n", wide2char (*mime_type), *len);
 			debug_write (error);
```
dbry: the printf "z" specifier is also a C99-ism

dbry:
Thanks again for this. It's funny, because last night I tried to compile on MSVC (which I do very rarely because it involves booting Windows) and saw that _I_ had violated the mid-block declaration recently. But I would not say that I am _targeting_ C99, but as long as I'm using MSVC for the official Windows builds I'll try to keep it working there (without warnings). Once I move to a newer MSVC or mingw32 for Windows builds, then that won't matter.

When we did the Travis-CI the Windows build came up, but yeah, it would break my stuff. At some point I'll update to a newer MSVC or just use mingw32 (although I currently have a weird issue where the mingw32 builds run fine on Windows but don't run correctly on Wine, which scares me a little).

Cheers!

PR #47, opened 2018-09-20
Optimize wavpack.h

evpobr:
The goal is to remove duplicated code from `wavpack.h` & `wavpack_local.h`. `WavPackContext` is still opaque pointer for library users, but for private use it is typedef to internal struct.

* Include `wavpack.h` to `wavpack_local.h`
* `WavPackContext` is now typedef of struct `WavPackContext` in
`wavpack_local.h`
* Remove duplicates from `wavpack_local.h`

### DIFF
--- a/src/wavpack_local.h
+++ b/src/wavpack_local.h
@@ -11,6 +11,8 @@
 #ifndef WAVPACK_LOCAL_H
 #define WAVPACK_LOCAL_H
 
+#include "wavpack.h"
+
 #if defined(_WIN32)
 #define strdup(x) _strdup(x)
 #define FASTCALL __fastcall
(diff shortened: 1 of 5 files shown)


src/wavpack_local.h
```diff
@@ -97,121 +99,8 @@ typedef struct {
...
-typedef struct {
-    char ckID [4];
-    uint32_t ckSize;
-    int16_t version;
-    unsigned char block_index_u8;
-    unsigned char total_samples_u8;
-    uint32_t total_samples, block_index, block_samples, flags, crc;
-} WavpackHeader;
-
-#define WavpackHeaderFormat "4LS2LLLLL"
```
jrlanglois: Just passing by with curiosity - why the removal of this macro and the many below? I wouldn't call that an optimisation - they'll get stripped away by the compiler.

dbry: I agree with and like this change and was going to merge it soon (I've been overwhelmed with day job recently).

However I just read your comment about the major ABI bump, which is kind of a show-stopper for me. Are you sure about that? The WavpackContext pointer gets changed from a void pointer to an opaque pointer, but in either case the user of libwavpack treats it exactly the same (i.e. they can only check it for NULL and pass it back to the library, no dereferencing of any kind). I'm sure that binaries would continue to work unmodified, but I'm sure that from some viewpoint the type did change even if they're equivalent. What do you think?

I'm sure it would also be possible to make an equivalent change without modifying wavpack.h at all, but it would be a little ugly and I like the additional type checking that this provides users of libwavpack.

dbry: Or is it an issue that the binaries _would_ actually break because of some sort of name-mangling and/or linker type-checking going on under the hood that I am blissfully ignorant of? Guess I could try it...at least on Linux.

evpobr:
It's better to test, maybe I'm wrong here. If so, ABI version could be untouched.

dbry:
I have decided to go ahead with this. The definition of WavpackContext does not change for the library code. What gets exported from `wavpack.h` changes from a void typedef to an opaque struct typedef, which adds type checking to the API and _should_ be equivalent from an ABI standpoint.

If this becomes a problem we can always change the definition in `wavpack.h` to:
```
#ifdef WAVPACK_LOCAL_H
typedef struct WavpackContext WavpackContext;
#else
typedef void WavpackContext;
#endif
```
so that everything's identical and we still eliminate the duplicated stuff.

Thanks for doing this!

Issue #54, opened 2018-11-29
AddressSanitizer: heap-buffer-overflow (Read OOB) at src/open_utils.c:1215


HongxuChen:
As of 6673775, when running `wvunpack $FILE`, it may result in an read out-of-bound error on variable `dp` inside `WavpackVerifySingleBlock` (blind decode mode is ok).
```
=================================================================                                                                                                                                                                              
==10571==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x606000000054 at pc 0x7fc8ab060e41 bp 0x7fff4f2a4b30 sp 0x7fff4f2a4b28                                                                                                      
READ of size 1 at 0x606000000054 thread T0                                                                                                                                                                                                     
    #0 0x7fc8ab060e40 in WavpackVerifySingleBlock /home/hongxu/FOT/WavPack/src/open_utils.c:1215:19                                                                                                                                            
    #1 0x7fc8ab05d9e7 in WavpackOpenFileInputEx64 /home/hongxu/FOT/WavPack/src/open_utils.c:124:14                                                                                                                                             
    #2 0x7fc8ab06b401 in WavpackOpenFileInput /home/hongxu/FOT/WavPack/src/open_filename.c:261:12                                                                                                                                              
    #3 0x517a4c in unpack_file /home/hongxu/FOT/WavPack/cli/wvunpack.c:1049:11                                                                                                                                                                 
    #4 0x51668b in main /home/hongxu/FOT/WavPack/cli/wvunpack.c:765:22                                                                                                                                                                         
    #5 0x7fc8a9c05b96 in __libc_start_main /build/glibc-OTsEL5/glibc-2.27/csu/../csu/libc-start.c:310                                                                                                                                          
    #6 0x41b419 in _start (/home/hongxu/FOT/WavPack/install/bin/wvunpack+0x41b419)                                                                                                                                                             
                                                                                                                                                                                                                                               
0x606000000054 is located 0 bytes to the right of 52-byte region [0x606000000020,0x606000000054)                                                                                                                                               
allocated by thread T0 here:                                                                                                                                                                                                                   
    #0 0x4db2d0 in __interceptor_malloc (/home/hongxu/FOT/WavPack/install/bin/wvunpack+0x4db2d0)                                                                                                                                               
    #1 0x7fc8ab05d612 in WavpackOpenFileInputEx64 /home/hongxu/FOT/WavPack/src/open_utils.c:111:43                                                                                                                                             
[...]

Issue #53, opened 2018-11-26
Infinite loop when block_samples==0 using wavpack


HongxuChen:
As to 6673775, wavpack might trigger an infinite loop at pack_utils.c:514 due to mis-calculation of `wpc->block_samples`.
wavpack --blocksize=128 -h $FILE -o /tmp/test.wv -y

example input wav files:
[wavpack_hangs.zip](https://github.com/dbry/WavPack/files/2616193/wavpack_hangs.zip)

dbry:
Thanks for reporting this! Closing as fixed.

carnil:
CVE-2018-19840 has been assigned for this issue.

PR #55, opened 2018-11-29
#53 #54 Ugly hack to only add the register keyword when a C compiler is present.

jrlanglois:
To help fix:
* #53 
* #54


dbry:
I don't this is _that_ ugly, but I don't think it's needed. As was mentioned, modern compilers probably don't even use the `register` keyword, and it didn't really have anything to do with those issues (they just happened to be filed against that commit, but have been there forever).

PR #57, opened 2018-12-08
x64 ASM support for midipix

Redfoxmoon3:
This adds support for building on midipix with the manually written ASM routines, only 64-bit version for now, as 32-bit doesn't compile yet, so I can't test it...


dbry:
Thanks! I had never even heard of the midipix project...I need to try that out.

PR #58, opened 2018-12-09
cli: Fix build with slibtool.

orbea:
The `.libs` directory is an internal libtool directory which build projects should not directly use. With libtool it will happen to work because it parses the actual file, but with other implementations such as slibtool it will fail because it relies a `.deps` file found in the same directory as the `.la` file which will not be found.

Instead of using the the symlinked `.la` file in the `.libs` directory its better to use the file the symlink points to and this will allow greater portability for building WavPack.

For reference here is the current build error with slibtool:
```
rdlibtool --tag=CC --mode=link gcc -I../include -g -O2 -o wavpack wavpack-wavpack.o wavpack-riff.o wavpack-wave64.o wavpack-caff.o wavpack-dsdiff.o wavpack-dsf.o wavpack-utils.o wavpack-md5.o wavpack-import_id3.o ../src/.libs/libwavpack.la -lm -lcrypto

rdlibtool: error logged in slbt_get_deps_meta(), line 125: No such file or directory.
rdlibtool: < returned to > slbt_exec_link_create_executable(), line 1463.
rdlibtool: < returned to > slbt_exec_link(), line 1845.
make[1]: *** [Makefile:509: wavpack] Error 2
make[1]: Leaving directory '/tmp/WavPack/cli'
make: *** [Makefile:457: all-recursive] Error 1
```
And the slibtool repo and documentation can be found here.
https://git.midipix.org/cgit.cgi/slibtool/

### DIFF
--- a/cli/Makefile.am
+++ b/cli/Makefile.am
@@ -8,7 +8,7 @@ wavpack_CFLAGS = $(AM_CFLAGS) $(ICONV_CFLAGS) -I$(top_srcdir)/include
 if ENABLE_RPATH
 wavpack_LDFLAGS = -rpath $(libdir)
 endif
-wavpack_LDADD = $(AM_LDADD) $(top_builddir)/src/.libs/libwavpack.la $(LIBM) $(ICONV_LIBS)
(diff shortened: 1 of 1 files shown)


dbry:
Thanks for this! I'm clueless about autotools and make, so any fixes there are very welcome.

PR #59, opened 2018-12-17
Add CMake project

evpobr:


dbry:
I ran into an issue when I tried to build with cmake. I got this:
```
CMake Warning at CMakeLists.txt:73 (find_package):
  By not providing "FindIconv.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Iconv", but
  CMake did not find one.

  Could not find a package configuration file provided by "Iconv" with any of
  the following names:

    IconvConfig.cmake
    iconv-config.cmake

  Add the installation prefix of "Iconv" to CMAKE_PREFIX_PATH or set
  "Iconv_DIR" to a directory containing one of the above files.  If "Iconv"
  provides a separate development package or SDK, be sure it has been
  installed.
```
I found the `FindIconv.cmake` file in `cmake/modules` but it wasn't being added to the `CMAKE_MODULE_PATH` because my cmake is 3.10 and that's only included for cmake _less than_ 3.10. I changed the line to 3.11 and everything worked fine, but I have no idea if that's the correct solution. Any ideas?

Other than that, everything I tried worked great. Thanks again!

evpobr:
> I found the `FindIconv.cmake` file in `cmake/modules` but it wasn't being added to the `CMAKE_MODULE_PATH` because my cmake is 3.10 and that's only included for cmake _less than_ 3.10. I changed the line to 3.11 and everything worked fine, but I have no idea if that's the correct solution. Any ideas?

Yes, it was my mistake. `FindIconv` module is shipped with CMake starting from 3.11.

dbry:
okay, thanks, I'll go ahead and just change the check to 3.11

PR #60, opened 2018-12-20
restrict _BitScanForward to msvc

tico-tico:
This should eliminate the _implicit declaration_ warning during MinGW build. MinGW has the `__builtin_ctz` function and defines `_WIN32` too. By some unknown magic it produces working executable even without the change, but with the change we get rid of the warning and the produced executable is a little bit smaller.

BTW here
https://github.com/dbry/WavPack/blob/a07f314ae89092362211a6f982e27c83febec086/src/read_words.c#L32

you restrict `_BitScanForward` only to win64 not sure why, i think it should work for win32 too (haven't tested though).

### DIFF
--- a/src/read_words.c
+++ b/src/read_words.c
@@ -124,7 +124,7 @@ int32_t FASTCALL get_word (WavpackStream *wps, int chan, int32_t *correction)
             wps->wvbits.bc += sizeof (*(wps->wvbits.ptr)) * 8;
         }
 
-#ifdef _WIN32
+#ifdef _MSC_VER
         { unsigned long res; _BitScanForward (&res, (unsigned long)~wps->wvbits.sr); ones_count = (uint32_t) res; }
 #else
         ones_count = __builtin_ctz (~wps->wvbits.sr);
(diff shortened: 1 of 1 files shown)


tico-tico:
> Confirm this warning. I guess `HAVE___BUILTIN_CTZ` is enough.

I don't understand what you are sayin. Are you talking about my patch or _line 32_? I guess the _line 32_ should be something like `#if defined (HAVE___BUILTIN_CTZ) || defined (_MSC_VER) ` to allow the __BitScanForward_ intrinsics in MSVC x32. But i'm not 100% sure since I can't test it.

dbry:
I think he just meant that `HAVE__BUILTIN_CTZ` will work with MinGW and I have also verified this warning (using `-Wall`). It could also be resolved by including `<intrin.h>`, but I think your solution is better and will merge this PR.

BTW, the reason I restricted `_BitScanForward` to win64 is because on win32 my `USE_NEXT8` optimization was faster (at least at that time). I'll check this again before releasing anything, but they were probably pretty close.

Thanks for the patch!

Issue #21, opened 2017-07-11
FLAC comparison


svnpenn:
This is almost certainly off topic, but as I currently use FLAC and I just learned that WavPack exists, it is really the only question that I care about. Does WavPack have any speed/size comparisons out there? I found this one:

http://fmedia.firmdev.com/audio-formats

which shows WavPack with slightly better compress, but getting destroyed in the decode time. Also it shows that WavPack cant even encode?

dbry:
Well, WavPack definitely encodes. That comparison only shows encoding speeds for codecs that fmedia can encode itself. A good rule of thumb is that WavPack, Monkey's Audio (APE), and OptimFROG are all symmetrical, so they roughly encode and decode at the same rate (unlike FLAC, as you can see from the table). Here's another comparison of the major lossless compressors: 

[http://wiki.hydrogenaud.io/index.php?title=Lossless_comparison](url)

I would say that if FLAC does exactly what you want, then there's no reason to consider alternatives. It's a great format and certainly the most widely supported, and any differences in compression with other formats are going to very small.

People generally turn to WavPack because it has a lot of features that FLAC does not have. For example, it handles 32-bit data (integer and float), it handles DSD audio, and it also has lossy and hybrid modes (PCM only) that might be useful in some situations.

ValZapod:
> it handles 32-bit data (integer and float)

Flac can also do 32 bit. At least one type of that.

Artoria2e5:
Flac does int32 and recently DSD (with no guarantee of losslessness). Is WavPack lossless with DSD and float32?

PR #77, opened 2019-11-24
Fix Bitstream alignment problem

srtate:
While doing a CVE study, I was looking at CVE-2018-10536 in wavpack and discovered another issue. The overflow in that CVE is when the putbits macro (wavpack_local.h) is used, but examining the macro it appears that a check is included so that it will not write past the end of the buffer. In particular, after advancing bs->ptr, it checks if bs->ptr==bs->end. Clearly this wasn't working properly because of the buffer overflow.

A little investigation showed that this is an alignment issue. The buffer is allocated as a block of bytes, but then cast to a (uint16_t *) to initialize bs->ptr. This then is stepped by 2's, since 16-bits at a time are added to the buffer. However, the bs->end pointer might not be aligned the same way, so bs->ptr can skip right past it (the equality test never notices).

A small test case shows this status for the bitstream object:

$3 = {buf = 0x5555557753ee, end = 0x5555557758df, ptr = 0x5555557753ee, 
  wrap = 0x7ffff7bb2af7 <bs_write>, error = 0, bc = 0, sr = 0}

Since "end" is odd and "ptr" is even and advanced by 2's, the two will never be equal, and ptr will advance right past end.

The proposed fix rounds down the "end" pointer to the proper alignment when the Bitstream is initialized (two places). The code is a little ugly, but is general purpose and should always work.

The buffer end test is really a backup check, and the rest of the code is designed so that the buffer should never fill up. However, it's a good backup test for safety, so it's important that it work correctly!

### DIFF
--- a/src/open_utils.c
+++ b/src/open_utils.c
@@ -809,7 +809,7 @@ static void bs_open_read (Bitstream *bs, void *buffer_start, void *buffer_end)
 {
     bs->error = bs->sr = bs->bc = 0;
     bs->ptr = ((bs->buf = (uint16_t *)buffer_start) - 1);
-    bs->end = (uint16_t *)buffer_end;
+    bs->end = (void *)(sizeof(*bs->ptr)*((uintptr_t)buffer_end/sizeof(*bs->ptr)));
     bs->wrap = bs_read;
 }
 
--- a/src/pack.c
+++ b/src/pack.c
@@ -1685,7 +1685,7 @@ static void bs_open_write (Bitstream *bs, void *buffer_start, void *buffer_end)
 {
     bs->error = bs->sr = bs->bc = 0;
     bs->ptr = bs->buf = buffer_start;
-    bs->end = buffer_end;
+    bs->end = (void *)(sizeof(*bs->ptr)*((uintptr_t)buffer_end/sizeof(*bs->ptr)));
     bs->wrap = bs_write;
 }
 


dbry:
Thanks so much for taking the time to figure this out, and provide a patch!

To mitigate the vulnerability I originally fixed the input format parsers to not request erroneous configurations, but you're right that the library should be able to handle these types of errors without crashing.

However, you're also right that your patch, while clever, is a little ugly and I think I'd rather prevent this situation higher up in the chain. You see, all WavPack metadata blocks (including bitstreams) are meant to be aligned on 2-byte boundaries, so a better solution is to make sure that condition holds everywhere. In other words, if your solution prevents an overflow that's great, but it also means there's something else very wrong somewhere else and we're just hiding it.

In the test cases that crashed, the actual issue preventing the overflow detection was that `max_blocksize` was odd, so that's an easy fix.

In the `bs_open_read()` case, if you look just above each of the three calls to that function, a check is made to prevent an odd buffer size from being requested, so that one is already safe.

Finally, I am adding some additional sanity checks at configuration so that someone encountering these issues will get more information about what's wrong (and provide _another_ preventative measure for the overflow). Please see my latest two commits for details.

Thanks again!

srtate:
Yes - making sure the block size is a multiple of 2 is a better and cleaner solution in the current codebase. It would break if the underlying I/O size (for bit I/O buffering) were increased to 32-bit chunks rather than 16-bit, but whether that would ever happen is beyond anything I know about this project.

dbry:
I think doing that I/O in 32-bit chunks _might_ be faster than 16-bit, but I have no plans to attempt that at this point (I'm not sure the improvement with 16-bit has been worth the trouble).

BTW, I'm a little curious as to why you were looking at this CVE as part of your study. Are you looking for something exploitable?

srtate:
As to why we were looking at this CVE: Nothing deep there. We're doing careful walkthroughs of buffer error (overflow and out-of-bounds reads) CVEs that were patched in Ubuntu 18.04 in its first year of release. We're looking for patterns to help improve static analysis tools. This CVE was #21 in our random sample. Incidentally, we also studied CVE-2018-10538 and CVE-2018-10539. There are 3 other wavpack cves that are on the list, but we probably won't get to those since we have a good enough random sample of vulnerabilities.

We definitely found some significant patterns that we believe we can catch with static analysis. More details on that will have to wait though!

PR #78, opened 2019-12-18
configure.ac: fix --disable-man

ffontaine:
$enableval should be used to know if the user has passed --enable-man or
--disable-man

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


dbry:
Thanks!

PR #87, opened 2020-11-10
fix memory leaks.

ihsinme:
the function has the ability to exit without first clearing the allocated memory.

### DIFF
--- a/cli/wavpack.c
+++ b/cli/wavpack.c
@@ -1737,6 +1737,7 @@ static int pack_file (char *infilename, char *outfilename, char *out2filename, c
         error_line ("can't create file %s!", use_tempfiles ? outfilename_temp : outfilename);
         DoCloseHandle (infile);
         WavpackCloseFile (wpc);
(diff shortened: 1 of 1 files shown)


dbry:
Thanks for this, good catch! I have some local changes I'm working on so I'll leave this up here for a bit until I can merge it cleanly.

ihsinme:
is there any news on this PR?

dbry:
Hi. I took a look at this and I don't understand why you are freeing `out2filename_temp` but not `outfilename_temp` also?

It seems like the same change would apply to that.

If you fix that I'll merge this in.

Thanks!

ihsinme:
yes, at first glance, you are right. I will certainly fix this shortly.

zodf0055980:
You forget add `;` in wavpack.c:1830

ihsinme:
fix in https://github.com/dbry/WavPack/pull/89  
I apologize for this mistake.

PR #88, opened 2020-12-14
fixing a possible memory leak.

ihsinme:
the form of using the realloc function can lead to the loss of a pointer to previously allocated memory.

the suggested fixes will eliminate this issue.

### DIFF
--- a/src/open_utils.c
+++ b/src/open_utils.c
@@ -656,10 +656,16 @@ static int read_sample_rate (WavpackContext *wpc, WavpackMetadata *wpmd)
 
 static int read_wrapper_data (WavpackContext *wpc, WavpackMetadata *wpmd)
 {
+    unsigned char *tmpbuf;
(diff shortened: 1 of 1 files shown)


dbry:
Hi, and thanks for pointing this out!

I agree that this is technically incorrect, but I feel that this is actually a non-issue. If a small allocation like that fails then there is something seriously wrong with the system and the process is likely to be shut down very soon (which will free everything). I don't think the extra code to handle this is worth it, especially considering that out-of-memory is not handled gracefully by the library anyway and there are many other instances of this exact same pattern.

ihsinme:
yes you are to a degree right. but I researched 33 similar places in your code except 3, they all immediately lead to an exception and really have no effect on the leak. however, for example in the corrected file open_utils.c, in 2 places the allocated memory is checked for NULL equality. which suggests that the author is ready for this situation. Moreover, these checks do not terminate the program.

so I propose to consider this PR.

PR #100, opened 2021-04-23
Update wavpackdll.vcxproj

c72578:
- The following export was missing in case of **`Debug|x64`**:
  `/export:WavpackGetEncodedNoise`
- Sort exports inside `AdditionalOptions` the same for all four
  combinations of `$(Configuration)|$(Platform)`:
    `Debug|Win32`
    `Debug|x64`
    `Release|Win32`
    `Release|x64`


dbry:
Thanks for catching this! Were you actually using that function??

c72578:
I came across the missing export of `WavpackGetEncodedNoise` in **`Debug|x64`** while updating WavPack to 5.4.0 in CUETools. However, the function `WavpackGetEncodedNoise` is not used in CUETools.

dbry:
Thanks!

PR #103, opened 2021-05-01
Clean up Autotools

SoapGentoo:
@dbry modernizing and simplifying the autotools part. There should be no functional changes.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,3 @@
-AUTOMAKE_OPTIONS = foreign
-
 if ENABLE_APPS
   MAYBE_APPS = cli
 endif
--- a/cli/Makefile.am
+++ b/cli/Makefile.am
@@ -4,21 +4,21 @@ wavpack_SOURCES = wavpack.c riff.c wave64.c caff.c dsdiff.c dsf.c utils.c md5.c 
 if WINDOWS_HOST
 wavpack_SOURCES += win32_unicode_support.c
 endif
-wavpack_CFLAGS = $(AM_CFLAGS) $(ICONV_CFLAGS) -I$(top_srcdir)/include
+wavpack_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/include
 if ENABLE_RPATH
 wavpack_LDFLAGS = -rpath $(libdir)
 endif
-wavpack_LDADD = $(AM_LDADD) $(top_builddir)/src/libwavpack.la $(LIBM) $(ICONV_LIBS)
+wavpack_LDADD = $(AM_LDADD) $(top_builddir)/src/libwavpack.la $(LIBM) $(LIBICONV)
 
 wvunpack_SOURCES = wvunpack.c riff_write.c wave64_write.c caff_write.c dsdiff_write.c dsf_write.c utils.c md5.c
 if WINDOWS_HOST
 wvunpack_SOURCES += win32_unicode_support.c
 endif
-wvunpack_CFLAGS = $(AM_CFLAGS) $(ICONV_CFLAGS) -I$(top_srcdir)/include
+wvunpack_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/include
 if ENABLE_RPATH
 wvunpack_LDFLAGS = -rpath $(libdir)
 endif
-wvunpack_LDADD = $(AM_LDADD) $(top_builddir)/src/libwavpack.la $(LIBM) $(ICONV_LIBS)
+wvunpack_LDADD = $(AM_LDADD) $(top_builddir)/src/libwavpack.la $(LIBM) $(LIBICONV)
 
 wvgain_SOURCES = wvgain.c utils.c
 if WINDOWS_HOST
@@ -38,7 +38,7 @@ wvtag_CFLAGS = $(AM_CFLAGS) -I$(top_srcdir)/include
 if ENABLE_RPATH
 wvtag_LDFLAGS = -rpath $(libdir)
 endif
-wvtag_LDADD = $(AM_LDADD) $(top_builddir)/src/libwavpack.la $(LIBM) $(ICONV_LIBS)
+wvtag_LDADD = $(AM_LDADD) $(top_builddir)/src/libwavpack.la $(LIBM) $(LIBICONV)
 
 check_PROGRAMS = wvtest
 wvtest_SOURCES = wvtest.c md5.c
--- a/configure.ac
+++ b/configure.ac
@@ -1,225 +1,175 @@
-# wavpack 5.4.0 configure.ac
+dnl wavpack 5.4.0 configure.ac
 
-AC_INIT(wavpack, 5.4.0, bryant@wavpack.com)
+AC_INIT([wavpack], [5.4.0], [bryant@wavpack.com])
 AC_CONFIG_SRCDIR([src/pack.c])
-AM_INIT_AUTOMAKE([serial-tests])
+AM_INIT_AUTOMAKE([-Wall 1.15 serial-tests subdir-objects foreign no-dist-gzip dist-xz])
 AM_MAINTAINER_MODE
 
 LIBWAVPACK_MAJOR=5
 LIBWAVPACK_MINOR=4
 LIBWAVPACK_MICRO=0
-LIBWAVPACK_VERSION_STRING=$LIBWAVPACK_MAJOR.$LIBWAVPACK_MINOR.$LIBWAVPACK_MICRO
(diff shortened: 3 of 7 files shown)


Makefile.am
```diff
@@ -1,5 +1,3 @@
-AUTOMAKE_OPTIONS = foreign
```
SoapGentoo, line 1: as stated in the last PR, modern Automake reads this from `configure.ac`, not Makefile.am

cli/Makefile.am
```diff
@@ -4,21 +4,21 @@ wavpack_SOURCES = wavpack.c riff.c wave64.c caff.c dsdiff.c dsf.c utils.c md5.c
 if WINDOWS_HOST
 wavpack_SOURCES += win32_unicode_support.c
 endif
-wavpack_CFLAGS = $(AM_CFLAGS) $(ICONV_CFLAGS) -I$(top_srcdir)/include
```
SoapGentoo, line 7: [`AM_ICONV` injects any necessary headers into `CPPFLAGS` instead](https://www.gnu.org/software/gettext/manual/html_node/AM_005fICONV.html)

configure.ac
```diff
@@ -1,225 +1,175 @@
...
 
-AC_SUBST(LT_CURRENT)
-AC_SUBST(LT_REVISION)
-AC_SUBST(LT_AGE)
+AC_SUBST([LT_CURRENT])
+AC_SUBST([LT_REVISION])
+AC_SUBST([LT_AGE])
 
-# Check for os version
-VERSION_OS=$(uname -s)
```
SoapGentoo, line 32: this is not portable to cross-compilation, because it will implicitly use the `CBUILD` os and not the `CHOST` one

configure.ac
```diff
@@ -1,225 +1,175 @@
...
-AC_DEFINE_UNQUOTED(VERSION_OS, "$VERSION_OS", [os version])
-
-# Check host so we can infer target CPU (and assembly optimizations)
+dnl Check for os version
 AC_CANONICAL_HOST
+AC_DEFINE_UNQUOTED([VERSION_OS], ["${host_os}"], [os version])
 
-# Check for processor characteristics
-AC_C_BIGENDIAN(AC_DEFINE([HIGHFIRST], [1], [big-endian machine]))
```
SoapGentoo, line 39: I checked the whole codebase and found no mention of either `HIGHFIRST` or `WORDS_BIGENDIAN`, so seems like dead code?

configure.ac
```diff
@@ -1,225 +1,175 @@
...
 AC_PROG_CC
-AC_LIBTOOL_WIN32_DLL
-AC_PROG_LIBTOOL
 AC_PROG_INSTALL
-AM_PROG_CC_C_O
+AM_PROG_AR
 AM_PROG_AS
 
-AC_HEADER_STDC
+LT_INIT([win32-dll disable-shared])
```
SoapGentoo, line 41: this just disabled shared libraries by default, users can always `--enable-static` if they wish

dbry:
Thanks so much for doing this! I am just out the door for the weekend, but I will try this all out first thing next week.

configure.ac
dbry: The code does check for `BYTE_ORDER` or `__BYTE_ORDER__` for endianess. It's less than optimum if those are not defined (slower code), but we will only generate bad code if they're defined and wrong (which doesn't seem likely). And yes, those you mentioned are not used anywhere.

SoapGentoo: You're right, but neither `AC_C_BIGENDIAN` nor its "success" block define `BYTE_ORDER` or `__BYTE_ORDER__` (both of which aren't portable), hence why I removed the macro.

dbry: I'm happy to do whatever's the correct method, but I do need to know whether I'm building for a big-endian or little-endian machine unless I want to get a significant performance hit. Is there a portable way to do that?

SoapGentoo: `AC_C_BIGENDIAN` is the correct way for that, you just aren't using its results to affect compilation. What about replacing `BYTE_ORDER` and friends with `WORDS_BIGENDIAN`, which is defined by `AC_C_BIGENDIAN` on big-endian machines? I can test this on our sparc machines.

dbry:
Travis and I are both having trouble with `AM_ICONV`. I updated Travis to focal (the latest) to get to Autotools 1.16, but still got failure. You can see them [here](https://travis-ci.org/github/dbry/WavPack). The Mac and Focal versions fail slightly differently, but they're both related to `AM_ICONV`.

SoapGentoo:
can you try installing [`gettext`](https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=blob;f=gettext-runtime/m4/iconv.m4;h=d0e61de2608d79d72ef78f25d98c9bdb3e07f777;hb=HEAD) too?

dbry:
```
gettext (GNU gettext-runtime) 0.19.8.1
Copyright (C) 1995-1997, 2000-2007 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Written by Ulrich Drepper.
```

SoapGentoo:
Where are you defining the dependencies you need to pull in for Homebrew in Travis?

dbry:
As far as I know, everything is in [this file](https://github.com/dbry/WavPack/blob/master/.travis.yml) and most of the Homebrew stuff is to build the man pages.

dbry:
This worked for me to build locally:
```
touch config.rpath
./autogen.sh
```

dbry:
Okay, I got the Travis Mac build fixed by upgrading from `xcode11` to `xcode12`. And my `touch` fix for Ubuntu worked on Travis too, but that's probably not the right fix, correct?

configure.ac
dbry: I actually think we're okay now that I've looked a little closer. The macros I check for are defined by the compiler (which is probably the most credible source) and I check for whether they're defined before I use them (so we can't error out). The purpose is to _only_ detect little endian, which I have some optimizations for, but the code will still work fine even if it's not detected (and the performance difference is not that great). No need for `AC_C_BIGENDIAN`.

SoapGentoo:
For the touch fix, try the following:

    cp /usr/share/gettext/config.rpath .

configure.ac
SoapGentoo: (as an aside, the code around the endianness dispatch includes undefined behavior: you're casting an `unsigned char*` to a `uint16_t*`, which violates strict aliasing rules, and with newer compiler versions/LTO/-O3/something might break. In general, pointer casting is a dangerous practice and should be avoided.)

dbry:
I guess that should go in `autogen.sh` since that's where everything needed to prep the environment should go, right? I don't imagine that the correct why to handle this is to put a recent copy of `config.rpath` in the repo, right? I made an attempt [here](https://github.com/dbry/WavPack/commit/953aeade1b9bf62f43f502a649e38a2f79caa2fc) and Travis is happy at last.

SoapGentoo:
Good question. `config.rpath` is a gettext file, so generally gets imported if you enable the whole gettext machinery (which you aren't). You can import it into your repo or add it in `autogen.sh`, I'd probably choose the latter if given the choice, because it is less likely to rot away in a repo and only gets copied into the root source dir when actually needed. Want me to add it in this PR?

dbry:
I created a test / scratch branch https://github.com/dbry/WavPack/compare/autotools-testing on top of your changes for triggering Travis runs. Not sure if you noticed, but a Travis run is triggered every PR or push to any branch, and you can see the latest run at https://travis-ci.org/github/dbry/WavPack.

I have pushed 6 times to that branch to get Travis happy again, including the `autogen.sh` change, so if you want to squash those changes and bring them into your PR, I can merge the whole thing together.

SoapGentoo:
@dbry all squashed into one commit

dbry:
I have tested this further and just about everything works just as expected. Thanks again for this!

I did have one question though. The required `automake` version is `1.16`. Is that really necessary, or did you just set that by default? The reason I ask is that `1.16` is fairly new and one of my development computers is on `Linux Mint 19.3` (barely a year old) and I really have no interest in upgrading it because it works fine and I don't want to break it. I rolled back the minimum requirement to `1.15` and everything worked great, so I'll use that for now, but I have no idea if that's actually acceptable.

SoapGentoo:
@dbry I've pulled it down to `1.15`. In general, because Automake is something that gets bootstrapped into the tarball, I consider requiring the latest version slightly less of a hassle than say, the latest GCC or Clang. That said, there's no feature in this PR (or one to come) that depends on 1.16, hence I've relaxed the lower bound to 1.15.

dbry:
Great, thanks!

PR #104, opened 2021-05-09
Make buildsystem non-recursive

SoapGentoo:
Hi @dbry, as promised, with this PR, I'm making the Automake-based buildsystem completely non-recursive.


dbry:
Thanks!

PR #105, opened 2021-05-09
iconv is not used in Windows builds, so don't require in configure

dbry:
@SoapGentoo I found a very minor issue with the new configure script where it would require iconv even for Windows builds (which use a native Windows API instead). This patch removes that dependency but I wanted to run it by you because I have no idea what I'm doing. Thanks.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,7 +55,15 @@ AC_ARG_ENABLE([apps],
   [AS_HELP_STRING([--disable-apps], [build only libwavpack (removes ICONV dependency)])])
 AM_CONDITIONAL([ENABLE_APPS], [test "x${enable_apps}" != "xno"])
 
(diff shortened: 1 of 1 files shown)


SoapGentoo:
yes looks good!

dbry:
Thanks!

Related to this, I just received notice that the Google Fuzz build is failing. Here's the log:

https://oss-fuzz-build-logs.storage.googleapis.com/log-7ef97b53-e5ff-42b3-b888-0fedba144743.txt

It seems to be that `AM_ICONV` is missing (we used to not actually use that macro, but implemented it ourselves in `configure.ac`).

Will this error be fixed if we just add `gettext` to the packages? I see that it's mentioned as suggested in the logs, but not brought it.

PR #106, opened 2021-05-13
Avoid obsolescent `test -a` constructs

SoapGentoo:
* POSIX calls the `-a` and `-o` operators obsolescent
  and strongly discourages their use, instead recommending
  chaining `test` calls instead:
  https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html


dbry:
Thanks!

PR #107, opened 2021-07-17
Cmake fixes 2021

evpobr:
* Use OpenSSL::Crypto target for CMake >= 3.4 instead of `${OPENSSL_INCLUDE_DIR}` and `${OPENSSL_CRYPTO_LIBRARY}`. It is not bug, just modern style CMake.
* `target_link_options` command requires CMake >= 3.13, version is too high. Now uses `target_link_options` if possible with fallback to `set_target_properties` for older CMake versions.

Both fixes come from my [Vcpkg](https://github.com/microsoft/vcpkg) port,  tested on Windows, Linux and macOS.


dbry:
Sorry for the delay. Thanks for this!

PR #111, opened 2022-01-03
Fix typos

rex4539:


dbry:
Hi, and thanks for these! My only issue is that the files in the `winamp` folders aren't mine and should not be changed.

If you resubmit the PR without those files I'll pull it in. Thanks!

rex4539:
Done :)

dbry:
Thanks! Funny, we all know how to spell `variants` now!  :grin:

PR #120, opened 2022-05-24
audition/cool_wv4.c: fix warnings from win64 builds

sezero:
```
audition/cool_wv4.c: In function 'FilterGetOptions':
audition/cool_wv4.c:772:5: warning: passing argument 4 of 'DialogBoxParamA' from incompatible pointer type
winuser.h:2240:29: note: expected 'DLGPROC' but argument is of type 'BOOL (*)(struct HWND__ *, UINT,  WPARAM,  LPARAM)'
audition/cool_wv4.c: In function 'WavPackDlgProc':
audition/cool_wv4.c:836:77: warning: cast from pointer to integer of different size
```

### DIFF
--- a/audition/cool_wv4.c
+++ b/audition/cool_wv4.c
@@ -673,7 +673,7 @@ void PASCAL CloseFilterInput (HANDLE hInput)
 }
 
 
-static BOOL CALLBACK WavPackDlgProc (HWND, UINT, WPARAM, LPARAM);
+static INT_PTR CALLBACK WavPackDlgProc (HWND, UINT, WPARAM, LPARAM);
 static int std_bitrate (int bitrate);
 
 // these variables are use to communicate with the dialog routines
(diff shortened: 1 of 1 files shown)


sezero:
I built them as 64 bit as part of a 64 bit build, when I noticed the build warnings.
I haven't tested, but there can _not_ be any kind of breakage because of this patch
because it only fixes win32 api usage and nothing else.

dbry:
Okay, I'll bring in the change (seems like a good idea). 

But currently the configuration on VS 2019 does _not_ build a 64-bit version of the Audition plugin. Did you override that for testing somehow, or do other versions of VS behave differently with these solution files. I am concerned because I don't want a 64-bit version of the Audition (or Winamp) plugin getting built and confusing someone (because they probably won't work)

Thanks!

sezero:
Ah, not MSVC:  I just built using MinGW.

dbry:
Ah, thanks, that explains it!

PR #125, opened 2022-07-11
cmake: Install header in same location

diizzyy:
CMake currently installs header file in root of includedir while a GNU Autotools installs in wavpack/.  Consumers expect to find the header in the latter so make sure to use the same path


dbry:
This is not an area I know much about, but I am used to the include file being in `/usr/local/include/wavpack/`

@evpobr should this just go in?

Thanks!

evpobr:
It looks like it. But I don't see the code that does it, only this:

https://github.com/dbry/WavPack/blob/b9efddc4a5fb63294cd8be63716d3d8da3efc137/Makefile.am#L101

evpobr:
If so, then it will be necessary to correct this line as well:

https://github.com/dbry/WavPack/blob/b9efddc4a5fb63294cd8be63716d3d8da3efc137/CMakeLists.txt#L204

About `CMAKE_INSTALL_INCLUDEDIR`: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

evpobr:
And this:

https://github.com/dbry/WavPack/blob/b9efddc4a5fb63294cd8be63716d3d8da3efc137/CMakeLists.txt#L652

diizzyy:
204 is also correct?
652 would break thinkgs as users expect to find the header in wavpack/wavpack.h not wavpack.h so changing that part would be wrong?

evpobr:
My mistake. If user must include `wavack/wavpack.h`, not `wavpack.h`, you don't need my changes.

dbry:
Okay @evpobr and @diizzyy, so this is good to go in as is? Thanks!

diizzyy:
Yes

evpobr:
Yes 😊

PR #126, opened 2022-07-13
autogen: Honour NOCONFIGURE=1

benjamb:
This otherwise gets in the way of being able to do out of tree builds when making use of autogen.sh.


dbry:
Thanks! You concur @SoapGentoo ?

SoapGentoo:
@dbry absolutely fine, we don't use the `autogen.sh` script anyways

PR #130, opened 2022-11-01
eliminate use of '&' address-of operator on arrays:

sezero:
fixes warnings from WatcomC (Open Watcom):
```
extra1.c(237): Warning! W115: &array may not produce intended result
extra1.c(556): Warning! W115: &array may not produce intended result
extra1.c(611): Warning! W115: &array may not produce intended result
extra2.c(396): Warning! W115: &array may not produce intended result
extra2.c(650): Warning! W115: &array may not produce intended result
extra2.c(746): Warning! W115: &array may not produce intended result
extra2.c(818): Warning! W115: &array may not produce intended result
tag_utils.c(529): Warning! W115: &array may not produce intended result
```

### DIFF
--- a/src/extra1.c
+++ b/src/extra1.c
@@ -234,7 +234,7 @@ static void recurse_mono (WavpackContext *wpc, WavpackExtraInfo *info, int depth
     if (branches < 1 || depth + 1 == info->nterms)
         branches = 1;
 
-    CLEAR (term_bits);
+    CLEARA (term_bits);
     samples = info->sampleptrs [depth];
     outsamples = info->sampleptrs [depth + 1];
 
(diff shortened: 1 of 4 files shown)


dbry:
Thanks for this! If you agree with my suggestion and update the PR, I'll bring it in.

sezero:
I'd update the PR, but what _is_ the suggestion?

dbry:
> I'd update the PR, but what _is_ the suggestion?

For the `src/pack_utils.c` change I think it would be better to change the `char zero [1] = { 0 };` to just `char zero = 0;` and leave the reference to it unchanged (i.e., `&zero`). Doesn't that make more sense?

sezero:
> For the `src/pack_utils.c` change I think it would be better to change the `char zero [1] = { 0 };` to just `char zero = 0;` and leave the reference to it unchanged (i.e., `&zero`). It doesn't need to be an array at all. Doesn't that make more sense?

Indeed. Updated and force-pushed the patch.

sezero:
If you are interested in watcom compiler support, I have initial support in my branch at https://github.com/sezero/WavPack/tree/watcom-0 (library only, no client.)

PR #131, opened 2022-11-05
mark x86 asm functions as __cdecl for Watcom compiler :

sezero:
`__cdecl` isn't default calling convention for Watcom.  This handles calling convention and also symbol decoration correctly.

### DIFF
--- a/src/wavpack_local.h
+++ b/src/wavpack_local.h
@@ -20,6 +20,12 @@
 #define FASTCALL
 #endif
 
+#if defined(__WATCOMC__) && defined(OPT_ASM_X86)
+#define ASMCALL __cdecl
+#else
+#define ASMCALL
+#endif
+
 #if defined(_WIN32) || \
     (defined(BYTE_ORDER) && defined(LITTLE_ENDIAN) && (BYTE_ORDER == LITTLE_ENDIAN)) || \
     (defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__))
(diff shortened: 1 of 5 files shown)


src/wavpack_local.h
```diff
@@ -20,6 +20,12 @@
 #define FASTCALL
 #endif
 
+#if defined(__WATCOMC__) && defined(OPT_ASM_X86)
```
sezero, line 23: I guess this could be generalized and `__WATCOMC__` condition can be dropped,
but I didn't want to poke the bear..

dbry: Actually I was thinking that myself, but `_cdecl` is ignored on ARM because it doesn't apply, and sometimes that can generate an error for "ignored attributes". Let me experiment.

sezero: With ARM, OPT_ASM_X86 should not be defined either..

dbry: Oh, I didn't read carefully. I was thinking that we could not add a macro at all and simply use `__cdecl` because it's ignored when not x86. While cleaner in some ways, that's liable to generate warnings sometimes. But yes, `__WATCOMC__` can certainly go.

sezero: >  But yes, `__WATCOMC__` can certainly go.

Well, it still is a problem: only MSVC (and Watcom) support that keyword.
MinGW (gcc targeting windows) does define it as `__attribute__((__cdecl__))`
so that one is OK too. But darwin-targeting gcc or clang doesn't define a
`__cdecl` macro. 

So I suggest that I keep  `__WATCOMC__` and really don't poke the bear:
what do you say?

dbry: Yeah, makes sense.

PR #133, opened 2022-11-07
win32, get_length: use GetFileSizeEx() instead of GetFileSize().

sezero:
This properly supports 64 bit file lengths. This obviously drops support for win9x and nt4, but the code already uses win32 apis available in win2k and newer.


sezero:
Actually no, forget this patch, not truly necessary. Closing.

PR #134, opened 2022-11-08
miscellaneous build / warning fixes:

sezero:
- cli/aiff.c: eliminate declaration-after-statement for c89 compilers
- cli/aiff.c: lose c99 vla and use a fixed [256] array for `compressionName`
- silence incompatible pointer warning for `GetFileSize()`.
- cli/utils.c: define `_WIN32_WINNT` as `0x0500` for `GetConsoleWindow()`.
- win32: don't define `fileno` as `_fileno` to avoid possible errors/warnings

### DIFF
--- a/cli/aiff.c
+++ b/cli/aiff.c
@@ -162,7 +162,7 @@ int ParseAiffHeaderConfig (FILE *infile, char *infilename, char *fourcc, Wavpack
                         int pstring_len = (unsigned char) common_chunk.compressionName [0];
 
                         if (pstring_len >= 1 && pstring_len <= chunk_header.ckSize - 23) {
-                            char compressionName [pstring_len + 1];
+                            char compressionName [256];
                             int i, j;
 
                             for (j = i = 0; i < pstring_len; ++i)
--- a/cli/aiff.c
+++ b/cli/aiff.c
@@ -242,6 +242,9 @@ int ParseAiffHeaderConfig (FILE *infile, char *infilename, char *fourcc, Wavpack
             }
         }
         else if (!strncmp (chunk_header.ckID, "SSND", 4)) {             // on the data chunk, get size and exit loop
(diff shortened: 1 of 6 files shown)


sezero:
If this goes in, watcom support is possible with only one FIXME (see: https://github.com/sezero/WavPack/commit/289938707a4914066bc5b329fff4755282d0e0b3 - should I create a PR for that after this?)

dbry:
Sorry, I really should have marked this as WIP. It's not at all tested and I haven't even built on MSVC yet (which is the intended toolchain for Windows). But okay, I'll work around this. Thanks!

dbry:
> If this goes in, watcom support is possible with only one FIXME (see: [sezero@2899387](https://github.com/sezero/WavPack/commit/289938707a4914066bc5b329fff4755282d0e0b3) - should I create a PR for that after this?)

I'm very curious about the Watcom support. Could you help me understand the purpose of this? I'm all for improving compatibility, but I'm also hesitant to promise that WavPack code will never utilize any improvements to the C language that are newer than 1989. And of course since I only use newer tools (having recently upgraded to MSVC 2019) it's always possible that I (or someone else) will break this accidentally and unless you're continuing to build nobody will ever know.

Finally, I assume that this does not target architectures with 16bit integers, right?

sezero:
> I'm very curious about the Watcom support. Could you help me understand the purpose of this?

My actual purpose was creating a usable os/2 version. I almost always work on and cross-compile from linux, and since the os/2 emx environment has never been ported to be used as a cross-toolchain, open watcom is my option.

> Finally, I assume that this does not target architectures with 16bit integers, right?

The compiler can do that, but I never intended that, nor have I ever tested in such a way. (And I very much doubt the library would be buildable in that way.)

I created PR #136, but I can close it or mark it as WIP. Your call.

PR #136, opened 2022-11-08
initial support for watcom compiler

sezero:
Notes:
1. FIXME: no 64 bit offset for `truncate_here()`
2. TODO: support os/2 for client programs?

The "`truncate_here` without 64 bit" is the only issue, but other than that
it works: tested by packing and unpacking a wav file under windows.


dbry:
What do you mean by the "client" programs? I see changes to `wavpack.c` and `wvunpack.c` so I assume those are okay, right?

sezero:
> What do you mean by the "client" programs? I see changes to `wavpack.c` and `wvunpack.c`

Yes, I meant clients of the library, i.e. wavpack, wvunpack, wvtag, wvgain ...

> so I assume those are okay, right?

Yes they are.

sezero:
Thanks.

Issue #137, opened 2022-11-09
truncate_here() question:


sezero:
As I understand it, `truncate_here` truncates the file at the _current_
position.  So: why do we need to use `_chsize_s` for windows from libc,
where we can simply call `SetEndOfFile` directly, i.e.:
```c
    int fd = fileno ((FILE *)id);
    HANDLE handle = (HANDLE) _get_osfhandle (fd);
    return (SetEndOfFile(handle) != 0) ? 0 : -1;
```
Is there some significance to calling `_chsize_s` ?

dbry:
No significance that I can think of, so `SetEndOfFile()` should work fine. Not sure if I just didn't find it before, or wanted more consistency with Linux (which I don't think has the equivalent). So you should use that for OS/2, but I think I'd rather keep Windows as it is ("if it ain't broke...")

BTW, that function is _only_ used in the case of editing tags and the file needs to be shortened because fields are deleted or edited smaller. If the `truncate_here()` callback doesn't exist (i.e. the pointer is set to `NULL`) then the code will zero-fill the file before the tag instead, which means the space is never reclaimed (which is generally not that big a deal). You might recall that this is the loop with the `zero` array that you recently edited!

sezero:
OK thanks.

dbry:
Thanks for the heads up on this! I didn't realize that Windows XP did not have `_chsize_s()` and I guess you didn't know I've been building winXP versions since 5.8.1.

Fortunately this is a pretty slim edge case, which is why I didn't catch it in testing. It works fine as is, unless an attempt is made to shorten or delete an APETAG on a file larger than 2GB *while actually running on XP*. There's demand for the winXP version for running compression tasks in the background (because it doesn't automatically try multithreading), but even those applications don't use  the `wvtag` application (which is the only way to trigger this issue from the command-line).

In any event, I guess you can fold this back into your SDL branch and don't have to maintain another diff. Actually, if I'd known you were maintaining a diff for this I would have just switched myself. Sorry.

sezero:
OK great that the patch is in.

> It works fine as is, unless an attempt is made to shorten or delete an APETAG on a file larger than 2GB _while actually running on XP_

Have you tested running on XP w/o the patch? (I haven't had a chance to do yet.) My guess was the exe (or dll) would fail running because of the missing export in msvcrt.dll ?

sezero:
[...]

PR #138, opened 2022-11-09
provide truncate_here() versions with 64 bit offset for watcom builds.

sezero:
c.f. https://github.com/dbry/WavPack/issues/137


dbry:
Cool.

PR #139, opened 2022-11-10
miscellaneous warning fixes

sezero:


dbry:
Hi @sezero , thanks for these! Could you let me know what generated these warnings (Watcom?) and would it be possible for you to send me the actual warning message texts? I'm a little curious, especially since some of them seem to me to be purely style.

I just finished a testing cycle and don't want to spend the time to bring these all in now and retest before doing a release candidate (especially because I take issue with some of them, and/or would fix them differently). But let's leave this PR up for reference.

sezero:
Only the FARPROC patch fixes watcom warnings. Others are gcc-generated (various versions, namely 4.9.4, 4.4.7, and 3.4.6.) Warnings themselves are exactly replicated in the commit messages, so I don't know what I should actually send to you?

dbry:
Oops, sorry, didn't click on the `...` to expand the commit messages! Yes, that gives me everything I need.

Funny, I'm using gcc 9.4.0 and don't get any of these (except the new parens for boolean mixed && and ||) even with `-Wall` (which I always check before a release, along with MSVC on warning level 3).

The one I was particularly curious about was the sequence point warnings. I can see where those _could_ be undefined if you, for example, do this:
`varA = function (++varB, varB + 10);`
because the evaluation order would change the result, but in my cases the evaluation order doesn't matter because the address of the variable doesn't change:
`varA = function (++varB, &varB);`
I've seen that warning when it's a real problem, but this one is obviously bogus.

I had also seen the possibly uninitialized `format` in `aiff.c` but I want to fix it by eliminating `format` completely (that's a hold over from WAV handling and using magic numbers like 1 and 3 is pretty ambiguous here).

Anyway, thanks again and let's leave this here until I get a chance to figure out what I'd like to do with all of them.

sezero:
Note that there is one certain bug fix in there, i.e. the success check of `SHGetSpecialFolderPathA` which is wrong in existing code.

sezero:
> Funny, I'm using gcc 9.4.0 and don't get any of these (except the new parens for boolean mixed && and ||) even with `-Wall` (which I always check before a release, along with MSVC on warning level 3).
> 
> The one I was particularly curious about was the sequence point warnings. I can see where those _could_ be undefined if you, for example, do this: `varA = function (++varB, varB + 10);` because the evaluation order would change the result, but in my cases the evaluation order doesn't matter because the address of the variable doesn't change: `varA = function (++varB, &varB);` I've seen that warning when it's a real problem, but this one is obviously bogus.

Note that the compiler warns about cases where `varB` is a pointer, so the address does change.

sezero:
> I see. Is that because `SHGetSpecialFolderPathA()` returns TRUE on success, and the SUCCEEDED macro tests for >= 0?

Yes.

>  Could you please put up a PR with just that file and I'll pull it in.

Done: https://github.com/dbry/WavPack/pull/140

dbry:
> > Funny, I'm using gcc 9.4.0 and don't get any of these (except the new parens for boolean mixed && and ||) even with `-Wall` (which I always check before a release, along with MSVC on warning level 3).
> > The one I was particularly curious about was the sequence point warnings. I can see where those _could_ be undefined if you, for example, do this: `varA = function (++varB, varB + 10);` because the evaluation order would change the result, but in my cases the evaluation order doesn't matter because the address of the variable doesn't change: `varA = function (++varB, &varB);` I've seen that warning when it's a real problem, but this one is obviously bogus.
> 
> Note that the compiler warns about cases where `varB` is a pointer, so the address does change.

I'm not sure I follow you. In my cases I'm passing the value of incremented pointer and the _address_ of the pointer, which should not change even when the pointer _value_ changes, right? The function overwrites the pointer value before returning, so I could also do this:
`set_console_title = strtol (argcp + 1, &argcp, 10);`
but I hesitate because I'm not entirely sure that it _always_ overwrites the value in every circumstance.

sezero:
> I'm not sure I follow you. In my cases I'm passing the value of incremented pointer and the _address_ of the pointer, 

Agh, you're right. (Sorry.)

dbry:
> > I'm not sure I follow you. In my cases I'm passing the value of incremented pointer and the _address_ of the pointer,
> 
> Agh, you're right. (Sorry.)

After this discussion I thought of another scenario (when I was supposed to be asleep). So consider this case:
```
char *argcp = argv;
...
set_console_title = strtol (++argcp, &argcp, 10);
```
The `strtol` function overwrites `argcp`, but what if the caller deferred writing back the incremented `argcp` until _after_ the function returns, thereby overwriting the value the function just stored? That's certainly not the intent. However, I checked, and the entry to a function _is_ a sequence point and so all side effects in the parameter list are guaranteed to be complete before the function is entered. Whew.

dbry:
I fixed most of these now. The sequence point ones I can't bring myself to change because it's definitely a bogus warning. The `cool_wv.c` warning was bogus too (as you said), but your change makes it cleaner. I hope that adding `ULL` will achieve the same thing as `UINT64_C()` (which I like better). Thanks!

sezero:
OK, thanks. The `UINT64_C` thing is only necessary if you are using older MSVC versions supporting not `ULL` but `ui64` (can't remember exact versions), but still OK I guess. I still pushed the sequence point patch here to this branch, at least for future reference.

Shall we close this PR?

dbry:
Naw, it can sit here for a while, at least until after the release. Thanks.

PR #140, opened 2022-11-11
utils.c: use prototyped function pointers instead of an opaque FARPROC.

sezero:
Also fixed the success check for SHGetSpecialFolderPathA() call.

Fixes the following warnings:

utils.c(405): Warning! W309: The function without prototyped parameters indirectly called
utils.c(411): Warning! W309: The function without prototyped parameters indirectly called
utils.c(424): Warning! W309: The function without prototyped parameters indirectly called

### DIFF
--- a/cli/utils.c
+++ b/cli/utils.c
@@ -382,12 +382,15 @@ extern int debug_logging_mode;
 
 #ifdef _WIN32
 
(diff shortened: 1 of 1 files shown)


dbry:
Thanks!

PR #141, opened 2022-11-11
enable clz and ctz optimizations for watcom builds.

sezero:
Uses `bsr` and `bsf` inline asm in watcom aux pragma syntax.

Question: these optimizations are not enabled in MSVC 32 bit
builds: is there a reason?

### DIFF
--- a/src/read_words.c
+++ b/src/read_words.c
@@ -31,6 +31,14 @@
 
 #if defined (HAVE___BUILTIN_CTZ) || defined (_WIN64)
 #define USE_CTZ_OPTIMIZATION    // use ctz intrinsic (or Windows equivalent) to count trailing ones
+#elif defined(__WATCOMC__) && defined(__386__)
+#define USE_CTZ_OPTIMIZATION
+extern __inline uint32_t _bsf_watcom (uint32_t);
+#pragma aux _bsf_watcom = \
+  "bsf eax, eax" \
+  parm [eax] nomemory \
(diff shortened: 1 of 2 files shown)


dbry:
> Uses `bsr` and `bsf` inline asm in watcom aux pragma syntax.
> 
> Question: these optimizations are not enabled in MSVC 32 bit builds: is there a reason?

I believe that every time I've tested it, the optimized version is slower. Haven't done that in some time though, and maybe not since I switched to MSVC 2019.

sezero:
> > I believe that every time I've tested it, the optimized version is slower.
> 
> Huh.. I see.

Interesting though, mingw and unix builds have them enabled by default provided that the builtins are available. Only MSVC x86 builds are the outliers.

Anyways, will this patch go into the upcoming release or you intend to delay it for later?

dbry:
> > > I believe that every time I've tested it, the optimized version is slower.
> > 
> > 
> > Huh.. I see.
> 
> Interesting though, mingw and unix builds have them enabled by default provided that the builtins are available. Only MSVC x86 builds are the outliers.

Yes, that sounds familiar. Only MSVC was slower using the builtins. But I may test again now that I'm using a newer VS and am on a newer CPU. Of course, I could test on a wide range of CPUs, but the difference was not very big in any case.
 
> 
> Anyways, will this patch go into the upcoming release or you intend to delay it for later?

I'll pull it in right now.

PR #147, opened 2022-11-17
Little CMake touches

madebr:
- make sure the created `.map` files are correct.
 Don't use `list(APPEND)`, but use `set(A "${A}newline")`.
 It doesn't matter much, but I made sure the generated map file are indented correctly.
- `check_c_source_runs` cannot be used when cross compiling.
[`check_c_source_compiles`](https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html) links an executable, so that can be used.
- Create a `WavPack::WavPack` target.
 With this change, wavpack can be used as:
  ```cmake
  option(MYGAME_WAVPACK_VENDORED "Use vendored WavPack")
  if(MYGAME_WAVPACK_VENDORED)
    add_subdirectory(vendored/wavpack EXCLUDE_FROM_ALL)
  else()
    find_package(wavpack REQUIRED)
  endif()
  target_link_libraries(mygame PRIVATE WavPack::WavPack)
  ```
- Build cooledit and winamp plug-ins as module.
  The main difference is no import library is generated, which is not needed as these libraries are supposed to be used with `LoadLibraryA` and not linked against at build time.


dbry:
Great, this works fine for me too! Will test the cooledit and winamp plugins next. Thanks @madebr !

madebr:
Glad I could be of help :)

PR #148, opened 2022-11-18
CMake + MSVC fix

madebr:
- Some MSVC configurations did not export `WavpackFloatNormalize`.
- In my last pr, I converted the library types to MODULE. Looks like one needs to set the `OUTPUT_NAME` property on them.
- I don't have MFC installed on my system, so I changed `#include <windows.h>` to `#include <windows.h>`.
  Everything keeps building. But I'm no expert in MFC stuff.

With these changes, running `dumpbin /exports` on the dll's, lng's or flt's give me no difference in exported symbols.

### DIFF
--- a/wavpackdll/wavpackdll.rc
+++ b/wavpackdll/wavpackdll.rc
@@ -7,7 +7,7 @@
 //
 // Generated from the TEXTINCLUDE 2 resource.
 //
-#include "afxres.h"
+#include <winresrc.h>
 
 /////////////////////////////////////////////////////////////////////////////
 #undef APSTUDIO_READONLY_SYMBOLS
@@ -34,7 +34,7 @@ END
 
 2 TEXTINCLUDE 
 BEGIN
-    "#include ""afxres.h""\r\n"
+    "#include <winresrc.rc>\r\n"
     "\0"
 END
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -543,7 +543,7 @@ if(WAVPACK_BUILD_WINAMP_PLUGIN)
     )
     target_link_libraries(in_wv_lng PRIVATE $<$<BOOL:${MSVC}>:-NOENTRY>)
     set_target_properties(in_wv_lng PROPERTIES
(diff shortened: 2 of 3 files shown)


wavpackdll/wavpackdll.rc
```diff
@@ -7,7 +7,7 @@
 //
 // Generated from the TEXTINCLUDE 2 resource.
 //
-#include "afxres.h"
+#include <windows.h>
```
sezero, line 10: `winresrc.h` wolud be better

wavpackdll/wavpackdll.rc
```diff
@@ -34,7 +34,7 @@ END
 
 2 TEXTINCLUDE 
 BEGIN
-    "#include ""afxres.h""\r\n"
+    "<windows.h>\r\n"
```
sezero, line 37: `winresrc.h` wolud be better

sezero: And it should be with the `#include ` directive

madebr: Yes of course, oopsy :)

madebr: Fixed in e8ac75f9ff9d4564117b6204323f3ed4845e910b

wavpackdll/wavpackdll.rc
madebr: Applied in e8ac75f9ff9d4564117b6204323f3ed4845e910b

dbry:
Thanks, guys!

wavpackdll/wavpackdll.rc
```diff
@@ -34,7 +34,7 @@ END
 
 2 TEXTINCLUDE 
 BEGIN
-    "#include ""afxres.h""\r\n"
+    "#include <winresrc.rc>\r\n"
```
c72578, line 37:
```suggestion
    "#include <winresrc.h>\r\n"
```

dbry: Thanks...fixed!

PR #149, opened 2022-11-19
CMake fixes for WinAMP

madebr:
With these fixes, I am able to let the WavPack Decoder plug-in appear in WinAMP.
Proof:
image

- By adding the objects of wavpack to `in_wv.dll` instead of linking to it, you don't need to copy the `wavpackdll.dll` to the winamp folder. Copying `wavpackdll.dll` to the `Plugins` folder wasn't enough, you had to copy it to the folder containing the winamp executable. By embedding wavpack in `in_wv.dll`, there are no more external dependencies.
- Link to static libgcc and libstdc++ to avoid needing the GNU standard library in a public location.

I suppose cooledit suffers from similar issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -536,7 +536,7 @@ if(WAVPACK_BUILD_WINAMP_PLUGIN)
         PREFIX ""
     )
     if(MINGW)
-        target_link_libraries(in_wv PRIVATE -mwindows)
+        target_link_libraries(in_wv PRIVATE -mwindows -static-libgcc -static-libstdc++)
     endif()
 
     add_library(in_wv_lng MODULE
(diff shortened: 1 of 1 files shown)


sezero:
>  By adding the objects of wavpack to `in_wv.dll` instead of linking to it, you don't need to copy the `wavpackdll.dll`

I _think_ @dbry is linking the plugin to static libwavpack, but I may be wrong.  However ...

>   Link to static libgcc and libstdc++ to avoid needing the GNU standard library in a public location.

... this may be the actual issue he has had. Otherwise, there is _no_ reason that in_wv plugin to fail loading.

madebr:
> > By adding the objects of wavpack to `in_wv.dll` instead of linking to it, you don't need to copy the `wavpackdll.dll`
> 
> I _think_ @dbry is linking the plugin to static libwavpack, but I may be wrong. However ...

This issue does not exist with a static wavpack library indeed, but this solution works with both shared and static wavpack.
So win win chicken dinner :)

dbry:
This loads into winamp (as you show) but the dialogs (Configure, About, file info...) don't work because the resource file `winamp/resource.rc` was not included for `in_wv`. Once I added that the dialogs worked too! I'll push that change. Thanks!

Issue #146, opened 2022-11-17
Using cmake with an alternate toolchain


dbry:
Since there have been numerous changes to the cmake stuff here, including the ability to build the winamp plugin for Window using mingw, I thought I would experiment with this (if for no other reason to test the generated winamp plugin and see if it actually works). I have mingw installed and use it to generate Windows executables all the time (by itself, or with Autotools to build WavPack), so I thought this would be fairly straightforward.

Initially I had no idea how to start so I Googled and found [this page](https://gist.github.com/peterspackman/8cf73f7f12ba270aa8192d6911972fe8). I tried that and got pretty far, but it fails with both mingw versions here:
```
Performing C SOURCE FILE Test COMPILER_SUPPORTS_SYMBOL_MAPS failed with the following compile output:
Change Dir: /home/david/WavPack/build/CMakeFiles/CMakeTmp

Run Build Command(s):/usr/bin/make cmTC_be63c/fast && /usr/bin/make -f CMakeFiles/cmTC_be63c.dir/build.make CMakeFiles/cmTC_be63c.dir/build
make[1]: Entering directory '/home/david/WavPack/build/CMakeFiles/CMakeTmp'
Building C object CMakeFiles/cmTC_be63c.dir/src.c.obj
/usr/bin/x86_64-w64-mingw32-gcc   -DCOMPILER_SUPPORTS_SYMBOL_MAPS -Wl,--version-script=/home/david/WavPack/build//CMakeFiles/conftest.map   -o CMakeFiles/cmTC_be63c.dir/src.c.obj   -c /home/david/WavPack/build/CMakeFiles/CMakeTmp/src.c
Linking C executable cmTC_be63c.exe
/usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_be63c.dir/link.txt --verbose=1
/usr/bin/cmake -E remove -f CMakeFiles/cmTC_be63c.dir/objects.a
/usr/bin/x86_64-w64-mingw32-ar cr CMakeFiles/cmTC_be63c.dir/objects.a @CMakeFiles/cmTC_be63c.dir/objects1.rsp
[...]

PR #150, opened 2022-11-21
misc patches to plugins:

sezero:
- add OpenWatcom makefile to build Win32 CoolEdit plugin.
- use prototyped function ptrs instead of an opaque FARPROC
  (same as https://github.com/dbry/WavPack/pull/140, but for winamp plugin.)
- winamp/wa_ipc.h: replace UNREFERENCED_PARAMETER(item) with (void)(item)
  win32 sdk defines `UNREFERENCED_PARAMETER(item)` as `(item)` which generates
  warnings with other compilers.
- winamp plugin, truncate_here(): watcom clib support

Not adding a watcom makefile for winamp plugin because the result is reported
to crash in ticket #146.

(Question: One curious thing in winamp plugin is that its reader passes pointers to `FILE*`
pointers instead just just the `FILE*` pointers themselves. Any specific reason??)

### DIFF
--- a/audition/Makefile.wat
+++ b/audition/Makefile.wat
@@ -1,0 +1,82 @@
+# OpenWatcom makefile to build CoolEdit plugin 'cool_wv4.flt' for Win32
+
+LIBNAME = cool_wv4
+
+DLLFILE = $(LIBNAME).flt
+LNKFILE = $(LIBNAME).lnk
+
+CC = wcc386
+RC = wrc
+
+WAVPACK_LIB = wavpack.lib
+
+CFLAGS = -bt=nt -d0 -zq -bm -5s -fp5 -fpi87 -sg -oeatxh -ei
+#CFLAGS+= -j
+# warnings:
+CFLAGS+= -wx
+# newer OpenWatcom versions enable W303 by default:
+CFLAGS+= -wcd=303
+# include paths:
+CFLAGS+= -I"$(%WATCOM)/h/nt" -I"$(%WATCOM)/h"
+CFLAGS+= -I"../include"
+# to build a dll:
+CFLAGS+= -bd
+RCFLAGS = -q -r -bt=nt -I"$(%WATCOM)/h/nt"
+
+SRCS = cool_wv4.c
+RCSRCS = wavpack.rc
+
+.extensions:
+.extensions: .lib .flt .obj .res .c .rc
+
+OBJS = $(SRCS:.c=.obj)
+RCOBJS= $(RCSRCS:.rc=.res)
+
+all: $(DLLFILE)
+
+$(DLLFILE): $(OBJS) $(RCOBJS) $(LNKFILE)
+  @echo * Link: $@
+  wlink @$(LNKFILE)
+
+$(LNKFILE):
+  @%create $@
+  @%append $@ SYSTEM nt_dll INITINSTANCE TERMINSTANCE
+  @%append $@ NAME $(DLLFILE)
+  @for %i in ($(OBJS)) do @%append $@ FILE %i
+  @%append $@ OPTION QUIET
+  @%append $@ OPTION RESOURCE=$(RCOBJS)
+  @%append $@ LIB $(WAVPACK_LIB)
+  @%append $@ export 'QueryCoolFilter'='_QueryCoolFilter@4'
+  @%append $@ export 'FilterUnderstandsFormat'='_FilterUnderstandsFormat@4'
+  @%append $@ export 'GetSuggestedSampleType'='_GetSuggestedSampleType@12'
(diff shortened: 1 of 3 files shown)


dbry:
> (Question: One curious thing in winamp plugin is that its reader passes pointers to `FILE*` pointers instead just just the `FILE*` pointers themselves. Any specific reason??)

Are you referring to, for example, [this line](https://github.com/dbry/WavPack/blob/5cf8a3922521aa5a279269ae20cd6a148e023fb8/audition/cool_wv4.c#L203) ?

Those aren't actually FILE pointers, but are pointers to a `write_id` struct, which includes a file pointer but also some fields that get updated by the `write_block()` callback.

Or maybe you were referring to something else?

sezero:
> Sorry, was looking in the wrong place!

The place is actually right :)

I'm referring to passing `&out->wv_file` instead of just `out->wv_file`
And the reader functions, rightfully, dereference their `void* id` param
to get the `FILE*` pointer like `*(FILE**)id` because of that instead of
just `(FILE*)id`.

dbry:
> > Sorry, was looking in the wrong place!
> 
> The place is actually right :)
> 
> I'm referring to passing `&out->wv_file` instead of just `out->wv_file` And the reader functions, rightfully, dereference their `void* id` param to get the `FILE*` pointer like `*(FILE**)id` because of that instead of just `(FILE*)id`.

Okay, so in the Cooledit filter it's _required_ to pass a pointer to the struct because there are things modified in there and, for instance, we keep track of the size of the first block written (which we sometimes have to read and rewrite).

As for the winamp plugin, it _does_ look like the extra level of indirection is unneeded. My guess is that at some point in the past that was a struct, or I planned to change it to a struct for some reason (debugging perhaps).

One thing I notice is that we are not using the `close` callback, and so maybe at some point in the past I was using that and setting the file pointer to NULL to indicate a closed file (which of course would be impossible if passing the file pointer itself). Note that having the extra indirection allows us to distinguish between a non-existent file (an unused wvc, for example) and a file that we've closed. Short answer: I have no idea!  :smiley:

sezero:
>  Short answer: I have no idea! 

Heh :)

dbry:
Tested by building both plugins on MSVC and verifying functionality. Thanks!

sezero:
One nit: In `truncate_here()` you probably want to return `-1` instead of 0
if file is NULL. I missed noticing that in the PR. Nothing big, just for better
correctness (I don't know if you check the result of truncate_here, but...)
```diff
diff --git a/winamp/in_wv.c b/winamp/in_wv.c
index dec5408..a07fb51 100644
--- a/winamp/in_wv.c
+++ b/winamp/in_wv.c
@@ -1298,7 +1298,7 @@ static int truncate_here (void *id)
         return _chsize_s (_fileno (file), curr_pos);
         #endif
     }
-    return 0;
+    return -1;
 }
 
 static WavpackStreamReader64 freader = {
```

dbry:
> One nit: In `truncate_here()` you probably want to return `-1` instead of 0 if file is NULL. I missed noticing that in the PR. Nothing big, just for better correctness (I don't know if you check the result of truncate_here, but...)
> 
> ```diff
> diff --git a/winamp/in_wv.c b/winamp/in_wv.c
> index dec5408..a07fb51 100644
> --- a/winamp/in_wv.c
> +++ b/winamp/in_wv.c
> @@ -1298,7 +1298,7 @@ static int truncate_here (void *id)
>          return _chsize_s (_fileno (file), curr_pos);
>          #endif
>      }
> -    return 0;
> +    return -1;
>  }
>  
>  static WavpackStreamReader64 freader = {
> ```

I'm not too concerned about that. I don't think that could _ever_ get called with NULL file, and if it did somehow then I'm not sure whether the correct result would be succeed or fail (other than not crashing). In the other places I have those reader callbacks I don't even check for NULL.

However, seeing that function reminded me of something that _will_ break. Since this is now being built on MINGW, I should probably put in [this fix](https://github.com/dbry/WavPack/commit/e9f817d703c33222047407b11f91348f6e74fddf). I also reported this to mingw  [here](https://sourceforge.net/p/mingw-w64/bugs/864/), but they don't seem too concerned (or that's not the best place for it).

sezero:
>  I also reported this to mingw [here](https://sourceforge.net/p/mingw-w64/bugs/864/), but they don't seem too concerned (or that's not the best place for it).

The bugs tracker unfortunately gets far much less attention than it deserves.
It's best that you report the issue to the `mingw-w64-public` mailing list:
https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/ (subscribing
to it is required before posting, _IIRC_.)

dbry:
And they'll probably want me to try it on the very latest (I'm on 9.3) which is not always easy when you're on the latest available for your distro.

PR #151, opened 2022-11-24
cmake: too many semicolons in libwavpack.map + escape string

madebr:
The version script was invalid on Linux.
It lacked semicolons because when writing the file, it treated FILE_CONTENTS as a list instead of a string.


dbry:
Thanks!

PR #154, opened 2023-01-04
manpages in mdoc(7)

janstary:
As agreed with @dbry, this replaces the man(7) manpages generated via XSLT from DocBook
with real, hand-written mdoc(7) manpages.

This means we can drop the entire DocBook / XSLT build dependency (way larger than wavpack itself),
there is nothing to "rebuild", and we have manpages everywhere.

Each of the new mdoc(7) manpages is about one half of the size of the old man(7),
and about one quarter of the old XML source; these can be removed as well.

Most importantly, this is a standard, readable, writeable,
manageable format providing semantic markup.

This branch also contains the relevant changes to the build system,
meaning the GNU auto*hell  and cmake. Please test and report any breakage;
I have tested these two on OpenBSD 7.2, FeeBSD 13.0, NetBSD 9.2, macOS 13.1 and Debian 11.3


dbry:
Okay, that looks great! There are a couple small syntax typos that I noticed on a quick look through the resulting pages, but I'll go more carefully once I pull it in and fix them (seems pretty straightforward to do).

Could you add your name to the manpage authors, unless you really don't want to?

Thanks!

janstary:
I'm not sure -- not that I oppose it, but I alway felt AUTHORS means authors of the software,
not the manpage; but if you preffer that, I'm adding it in another commit.

PR #156, opened 2023-01-10
Manpages

dbry:
@janstary , how about this?

Seems like the mdoc specs allow using `.Ss` to subdivide the `DESCRIPTION`, which I like (at least when there are lots of options). Checked with `mandoc -Tlint` and nothing new. Also your cmake change and several other minor tweaks.

Thanks!

### DIFF
--- a/man/wavpack.1
+++ b/man/wavpack.1
@@ -30,15 +30,16 @@ and in that case
 may be used to specify an alternate target directory.
 A filename of
 .Dq -
-specifies the standard input and output.
+specifies
+.Pa stdin
+or
+.Pa stdout .
 .Pp
 When transcoding from existing WavPack files,
 all tags are copied, and may be modified with additional args;
 unless an alternate output file or directory is specified,
 the source files are safely overwritten.
-.Pp
-The following input formats are supported:
-.Pp
+.Ss INPUT FORMATS
 .Bl -bullet -compact
 .It
 Microsoft RIFF, extension
@@ -64,8 +65,7 @@ Philips DSDIFF, extension
 Sony DSD Stream, extension
 .Dq .dsf
 .El
-.Pp
-The options are as follows:
+.Ss OPTIONS
 .Bl -tag -width Ds
 .It Fl a
 Adobe Audition (CoolEdit) mode for 32-bit floats
@@ -73,7 +73,7 @@ Adobe Audition (CoolEdit) mode for 32-bit floats
 Allow tag data up to 16 MB.
 Embedding > 1 MB is not recommended for portable devices
 and may not work with some programs, including older WavPack versions.
-.It Fl b Ar n
+.It Fl b Ns Ar n
 enable hybrid compression,
 .Ar n
 = 2.0 to 23.9 bits/sample, or
@@ -103,7 +103,8 @@ On by default in lossless mode and with the
 .Fl cc
 option.
 .It Fl d
-Delete source file if successful.
+delete source file if successful;
+.Sy use with caution!
 .It Fl f
 fast mode; introduces a compromise in compression ratio
 .It Fl -force-even-byte-depth
@@ -208,7 +209,7 @@ write program version to
 .Pa stdout
 .It Fl w Encoder
 write encoder metadata to APEv2 tag (e.g.,
-.Dq Encoder=WavPack 5.5.0 )
+.Dq Encoder=WavPack 5.6.0 )
 .It Fl w Settings
 write user settings metadata to APEv2 tag (e.g.,
 .Dq Settings=-hb384cx3 )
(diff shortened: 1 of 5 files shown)


man/wavpack.1
```diff
@@ -64,16 +65,15 @@ Philips DSDIFF, extension
...
+.Ss OPTIONS
 .Bl -tag -width Ds
 .It Fl a
 Adobe Audition (CoolEdit) mode for 32-bit floats
 .It Fl -allow-huge-tags
 Allow tag data up to 16 MB.
 Embedding > 1 MB is not recommended for portable devices
 and may not work with some programs, including older WavPack versions.
-.It Fl b Ar n
+.It Fl b Ns Ar n
```
janstary, line 76: Ah, wavpack doesn't use getopt(3) so this matters.

man/wavpack.1
```diff
@@ -103,7 +103,8 @@ On by default in lossless mode and with the
 .Fl cc
 option.
 .It Fl d
-Delete source file if successful.
+delete source file if successful;
```
janstary, line 106: In general, I preffer if the description is valid sentences,
including the capitalization; but these are nits regarding content
(as opposed to markup), which we can address later.

man/wavpack.1
```diff
@@ -103,7 +103,8 @@ On by default in lossless mode and with the
...
+.Sy use with caution!
```
janstary, line 107: I wouldn't mark this up at all, the exclamation mark is imho enough :-)
Perhaps Em (emphasis) would be more suitable than Sy (symbolic).

man/wavpack.1
```diff
@@ -231,7 +232,8 @@ extra encode processing, n = 0 to 6, default=1;
 -x1 to -x3 to choose best of predefined filters,
 -x4 to -x6 to generate custom filters (very slow!)
 .It Fl y
-yes to all warnings (use with caution!)
+yes to all warnings;
+.Sy use with caution!
```
janstary, line 236: same as above

man/wvtag.1
```diff
@@ -6,7 +6,7 @@
 .Nd manipulate wavpack metadata
 .Sh SYNOPSIS
 .Nm wvtag
-.Op Fl cdhlqvwxy
+.Op Fl options
```
janstary, line 9: I don't see how this is better: we know what the actual option letters are,
and it is customary to list them here, as in e.g. ls(1) or cat(1) or anything else.

man/wvunpack.1
```diff
@@ -172,18 +174,18 @@ about WavPack file to
 .Pa stdout
 .It Fl -skip=[-][ Ns Ar sample Ns | Ns Ar hh : Ns Ar mm : Ns Ar ss.ss ]
 start decoding at specified sample or time index, specifying a
-.Fl
+.Sy -
```
janstary, line 177: I's say `.Sq -` as it's a literal minus char.

man/wvunpack.1
```diff
@@ -172,18 +174,18 @@ about WavPack file to
...
 causes sample/time to be relative to EOF
 .It Fl t
 copy input file's time stamp to output file(s)
 .It Fl -until=[+|-][ Ns Ar sample Ns | Ns Ar hh : Ns Ar mm : Ns Ar ss.ss ]
 stop decoding at specified sample or time index, specifying a
-.Fl +
+.Sy +
```
janstary, line 183: imho Sq as above

man/wvunpack.1
```diff
@@ -172,18 +174,18 @@ about WavPack file to
...
 causes sample/time to be relative to
 .Fl -skip
 point,
 specifying a
-.Fl
+.Sy -
```
janstary, line 188: imho Sq as above

man/wvunpack.1
```diff
@@ -213,7 +215,8 @@ replaced with the extension from the binary tag source file (or
 .Dq txt
 for text tag).
 .It Fl y
-yes to overwrite warning (use with caution!)
+yes to overwrite warning;
+.Sy use with caution!
```
janstary, line 219: see above

janstary:
Comments inline, but generaly OK with me.
Thanks for checking and catching the errors.

janstary:
(BTW, these commint still amount to +38 -39 lines, so still good :-)

man/wvunpack.1
dbry: Yes, `.Sq` is a better fix. The `.Fl` was okay for '-' but not for the '+' where it displayed '-+'

man/wavpack.1
dbry: I'm happy to take it out if you feel strongly about it. But I believe that Sy (which the mdoc(7) page describes as for "importance or seriousness" and uses a warning as an example) is better than Em. Emphasis is used to _clarify_ a statement, which is _not_ the same as stressing its importance. Like I did there...  :smile_cat:

Besides, when I tried Em on my system I got a broken underline, which was ugly.

janstary: OK, no strong feelings.

man/wvunpack.1
janstary: The Fl is actualy wrong for even the `-`, as it's not a flag, but a literal minus; so your edit is right.

janstary:
Thanks again for the carefull check! I'm OK with these.

dbry:
@janstary thanks for the review!

PR #164, opened 2023-08-30
cmake: don't enable tests and programs when building wavpack as a subproject

### LINKED ISSUE #163: CMake default options
We are using WavPack as git submodule/directory in our project to build it statically as part of an "audiofile static library" for our [sfz](https://sfzformat.com) library project.

Just noticed in CMake configuration:

```
-- The following features have been enabled:

 * WAVPACK_ENABLE_DSD, Enable support for WavPack DSD files.
 * WAVPACK_ENABLE_THREADS, Enable support for threading in libwavpack.
 * WAVPACK_INSTALL_CMAKE_MODULE, Generate and install CMake package configuration module.
 * WAVPACK_ENABLE_ASM, Enable assembly optimizations.
 * WAVPACK_ENABLE_LIBCRYPTO, Use OpenSSL::Crypto library.
 * WAVPACK_BUILD_PROGRAMS, Build programs.
 * WAVPACK_INSTALL_DOCS, Install documentation.
 * WAVPACK_INSTALL_PKGCONFIG_MODULE, Generate and install wavpack.pc.
 ```
maybe the build programs and install xyz options could be disabled automatically if the project is not the top most (sub project), what do you think? Though I'm not sure about others like libcrypto and dsd.
```cmake
set(PROJECT_IS_MAIN OFF)
if(PROJECT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
    set(PROJECT_IS_MAIN ON)
endif()
cmake_dependent_option(WAVPACK_BUILD_PROGRAMS "Build programs" ON
    "(WIN32 OR Iconv_FOUND) AND PROJECT_IS_MAIN" OFF)
# ...
```
though it seems there are 2 problems:
from [cmake_dependent_option documentation](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html):
> New in version 3.22: Full [Condition Syntax](https://cmake.org/cmake/help/latest/command/if.html#condition-syntax) is now supported. See policy [CMP0127](https://cmake.org/cmake/help/latest/policy/CMP0127.html#policy:CMP0127).

Isn't `OR` (already there, despite my added `AND`) part of that full condition syntax, so that it requires at least CMake 3.22?

2nd problem:
`BUILD_TESTING` is used also in another library (Google Abseil). I don't think it's a good idea generic variable names like that, can you rename it to `WAVPACK_BUILD_TESTING`?

madebr:
This only changes the default when using wavpack as a subproject.
It is always possible to explicitly build the programs by configuring with `-DWAVPACK_BUILD_PROGRAMS=ON`.
I did not touch all options, as some are useful to have enabled by default.

@redtide
FIxes #163

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,12 @@ math(EXPR DYLIB_COMPATIBILITY_VERSION_MICRO "0")
 set(DYLIB_CURRENT_VERSION "${DYLIB_CURRENT_VERSION_MAJOR}.${DYLIB_CURRENT_VERSION_MINOR}.${DYLIB_CURRENT_VERSION_MICRO}")
 set(DYLIB_COMPATIBILITY_VERSION "${DYLIB_COMPATIBILITY_VERSION_MAJOR}.${DYLIB_COMPATIBILITY_VERSION_MINOR}.${DYLIB_COMPATIBILITY_VERSION_MICRO}")
 
+if(CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR)
+    set(WAVPACK_ROOTPROJECT ON)
+else()
+    set(WAVPACK_ROOTPROJECT OFF)
+endif()
+
 # Languages
 
 include(CheckLanguage)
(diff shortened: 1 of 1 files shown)


dbry:
Thanks!

Issue #116, opened 2022-03-14
Cannot find source file: wavpackdll/wavpackdll.rc


jonaski:
I'm getting the following error when compiling on Windows with MSVC (Visual Studio 2019).

```
C:\Data\Projects\strawberry\strawberry-dependencies\msvc\sources\wavpack-5.4.0\build>cmake .. -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="c:\msvc_x86_64" -DBUILD_SHARED_LIBS=ON -DBUILD_TESTING=OFF -DWAVPACK_BUILD_DOCS=OFF -DWAVPACK_BUILD_PROGRAMS=OFF -DWAVPACK_ENABLE_ASM=OFF -DWAVPACK_ENABLE_LEGACY=OFF -DWAVPACK_BUILD_WINAMP_PLUGIN=OFF -DWAVPACK_BUILD_COOLEDIT_PLUGIN=OFF
-- The C compiler identification is MSVC 19.29.30141.0
-- The CXX compiler identification is MSVC 19.29.30141.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for a ASM compiler
-- Looking for a ASM compiler - C:/Strawberry/c/bin/gcc.exe
-- The ASM compiler identification is GNU
-- Found assembler: C:/Strawberry/c/bin/gcc.exe
-- The ASM_MASM compiler identification is MSVC
-- Found assembler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30133/bin/Hostx64/x64/ml64.exe
-- Found OpenSSL: C:/msvc_x86_64/lib/libcrypto.lib (found version "3.0.1")
-- Looking for cos in m
[...]

PR #167, opened 2023-10-12
fix several build issues, add version res to autotools windows builds.

sezero:
- autotools: add version resource to libwavpack windows builds.
- autotools: link libcryto only to its users, not everything.
- autotools: don't link with pthread when targeting windows where it's
  not used and need not be available.
- cmake: fix openssl/libcrypto usage: it minked it to libwavpack where
  it isn't used and not the programs where it actually matters.
- cmake: bumped minimum required version to 3.5 to ease the use of ssl
  options.
- cmake: disable libcrypto by default so it matches autotools and it's
  noted that it breaks macOS builds.
- cmake: fix the purpose description of openssl package.
- cmake: only do find_package(Threads) for non-windows: otherwise it'd
  find pthreads which may be available in mingw toolchains.

Build tested using autotools and cmake targeting linux and windows (mingw)

Please review / test for any gotchas.

Cmake changes are least readable, at least for me, because of millions of
brackets, but builds were OK at my end. @madebr can verify.

P.S.: @dbry: If you want libcryto removed, I can change the patch for it.

P.S./2: I couldn't break the patch into smaller pieces doing one thing at
a time: they'd be inter-dependendent too much.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2...3.5)
+cmake_minimum_required(VERSION 3.5)
 
 project(WavPack VERSION 5.6.6)
 
@@ -96,7 +96,9 @@ if(HAVE_LIBM)
     list(APPEND CMAKE_REQUIRED_LIBRARIES m)
 endif()
 
-find_package(Threads)
+if(NOT WIN32)
+  find_package(Threads)
+endif()
 find_package(Iconv)
 find_package(OpenSSL)
 
@@ -147,7 +149,7 @@ endif()
 # Dependent options
 
 cmake_dependent_option(WAVPACK_ENABLE_ASM "Enable assembly optimizations" ON "HAVE_ASM OR HAVE_MASM" OFF)
-cmake_dependent_option(WAVPACK_ENABLE_LIBCRYPTO "Use OpenSSL::Crypto library" ON "OPENSSL_FOUND" OFF)
+cmake_dependent_option(WAVPACK_ENABLE_LIBCRYPTO "Use OpenSSL::Crypto library" OFF "OPENSSL_FOUND" OFF)
 cmake_dependent_option(WAVPACK_BUILD_PROGRAMS "Build programs" ${WAVPACK_ROOTPROJECT} "WIN32 OR Iconv_FOUND" OFF)
 cmake_dependent_option(WAVPACK_BUILD_COOLEDIT_PLUGIN "Build CoolEdit plugin" ${WAVPACK_ROOTPROJECT} "WIN32" OFF)
 cmake_dependent_option(WAVPACK_BUILD_WINAMP_PLUGIN "Build WinAmp plugin" ${WAVPACK_ROOTPROJECT} "WIN32" OFF)
@@ -204,26 +206,10 @@ target_include_directories(wavpack
 )
 target_link_libraries(wavpack
     PRIVATE
-    $<$<BOOL:${WAVPACK_ENABLE_THREADS}>:Threads::Threads>
+        $<$<AND:$<NOT:$<BOOL:${WIN32}>>,$<BOOL:${WAVPACK_ENABLE_THREADS}>>:Threads::Threads>
         $<$<BOOL:${HAVE_LIBM}>:m>
 )
 add_library(WavPack::WavPack ALIAS wavpack)
-
-if(CMAKE_VERSION VERSION_LESS 3.4)
-    target_include_directories(wavpack
-        PRIVATE
-            $<$<BOOL:${WAVPACK_ENABLE_LIBCRYPTO}>:${OPENSSL_INCLUDE_DIR}>
-    )
-    target_link_libraries(wavpack
-        PRIVATE
-            $<$<BOOL:${WAVPACK_ENABLE_LIBCRYPTO}>:${OPENSSL_CRYPTO_LIBRARY}>
-    )
-else()
-    target_link_libraries(wavpack
-        PRIVATE
-            $<$<BOOL:${WAVPACK_ENABLE_LIBCRYPTO}>:OpenSSL::Crypto>
-    )
-endif()
 
 target_compile_definitions(wavpack
     PRIVATE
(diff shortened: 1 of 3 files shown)


madebr:
> Cmake changes are least readable, at least for me, because of millions of brackets, but builds were OK at my end. @madebr can verify.

Yeah, CMake syntax will never win a beauty contest.
Your changes look good, and build fine on my machine.

dbry:
This looks good to me as well. I built/tested on both Pop!_OS and Raspberry Pi OS without issue.

I also built on both platforms with and without libcrypto and found the difference in performance non-existent on ARM (maybe there's no optimization). On x86_64 libcrypto improved the MD5 calculation speed by maybe 5%, but it was still a very small portion of the total time (< 2%).

If you would like to remove it, that would be fine with me. If you do, it would probably be better in a separate PR that does just that (after this one goes in?) and mention that the performance improvement was negligible (and certainly not worth the complexity).

Thanks so much!

sezero:
OK, I will remove libcrypto in a new p/r after this goes in.

sezero:
Just pushed a minor follow-up patch to this PR: wvtest relies on pthreads only and doesn't use win32 threads: Disabled it in autotools (with a FIXME note added), just like the cmake script already does.

sezero:
This PR is complete as it is now. Will post the new libcrypt-removal patch
after this gets in.

There is another build (+packaging) issue with cmake: When you configure
using cmake targeting windows from a release tarball, it fails like this:
```
-- Configuring done
CMake Error at CMakeLists.txt:550 (add_library):
  Cannot find source file:

    winamp/wavpack.rc

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx


CMake Error at CMakeLists.txt:530 (add_library):
  Cannot find source file:

    winamp/in_wv.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx


CMake Error at CMakeLists.txt:508 (add_library):
  Cannot find source file:

    audition/cool_wv4.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx


CMake Error at CMakeLists.txt:550 (add_library):
  No SOURCES given to target: in_wv_lng


CMake Error at CMakeLists.txt:530 (add_library):
  No SOURCES given to target: in_wv


CMake Error at CMakeLists.txt:508 (add_library):
  No SOURCES given to target: cool_wv4


CMake Generate step failed.  Build files cannot be regenerated correctly.
```

That's because you don't package certain windows-specific things such as
masm-specific asm, winamp and cooledit plugins, and MSVC project files.
Is that really by design? IMO, that shouldn't be, but it's just me. (I can
adjust autotools `make dist` to include missing stuff if you require..)

dbry:
> Just pushed a minor follow-up patch to this PR: wvtest relies on pthreads only and doesn't use win32 threads: Disabled it in autotools (with a FIXME note added), just like the cmake script already does.

My TODO list includes updating wvtest to use the threading macros. Shouldn't be too hard.

> That's because you don't package certain windows-specific things such as masm-specific asm, winamp and cooledit plugins, and MSVC project files. Is that really by design? IMO, that shouldn't be, but it's just me. (I can adjust autotools make dist to include missing stuff if you require..)

I guess it's by design. Sort of. I used to have a separate .zip file for building on Windows that included all the proper files, including the MSVC project files. Later, I got tired of creating and verifying that (manually) and just assume that Windows people will use the Git sources (which are also available as a .zip file on the [GitHub release page](https://github.com/dbry/WavPack/releases/tag/5.6.0)). I still think of tarballs as *nix (or MacOS) things and why would a Windows developer deal with that?

Last year you might remember that someone was having trouble (issue #116) because `wavpackdll.rc` was missing from the tarball, so I added that. But looking at that issue now I see that he had to disable the ASM, presumably because the MASM files were missing? So _maybe_ it makes sense to include those because it's just a few files and allows ASM builds on MSVC using just the tarball and cmake.

However, I think including the `winamp` and `Cool Edit` stuff is crazy. Those should probably be separate repos like the [DirectShow filter](https://github.com/dbry/WavPack-DirectShow) and [newer Audition plugins](https://github.com/dbry/amioWavpack) are. I think it's cool that we (mostly you and @madebr) were able to get them building and working under cmake/mingw, but I still don't think that they should be built by default, and certainly not the 64-bit versions _that aren't even a thing_ and can only cause problems if someone tries to actually use them.

sezero:
Created #168 for killing libcrypt option.

As for dist: Leaving the cooledit and winamp stuff out, would you be willing to accept something like the following or should I drop it?
```diff
diff --git a/Makefile.am b/Makefile.am
index 670cbfe..c6e33c2 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -14,7 +14,23 @@ EXTRA_DIST = \
 	CMakeLists.txt \
 	cmake/CheckCLinkerFlag.cmake \
 	cmake/modules/FindIconv.cmake \
-	cmake/TestLargeFiles.cmake
+	cmake/TestLargeFiles.cmake \
+	\
+	cli/Makefile.w32 \
+	src/Makefile.os2 \
+	src/Makefile.w32 \
+	src/pack_x64.asm \
+	src/pack_x86.asm \
+	src/unpack_x64.asm \
+	src/unpack_x86.asm \
+	\
+	src/libwavpack.vcxproj \
+	wavpackexe/wavpack.vcxproj \
+	wavpackdll/wavpackdll.vcxproj \
+	wvgainexe/wvgain.vcxproj \
+	wvunpackexe/wvunpack.vcxproj \
+	wvtagexe/wvtag.vcxproj \
+	wavpack.sln masm.rules
 
 MAINTAINERCLEANFILES = \
 	aclocal.m4 \
```

madebr:
> I guess it's by design. Sort of. I used to have a separate .zip file for building on Windows that included all the proper files, including the MSVC project files. Later, I got tired of creating and verifying that (manually) and just assume that Windows people will use the Git sources (which are also available as a .zip file on the [GitHub release page](https://github.com/dbry/WavPack/releases/tag/5.6.0)). I still think of tarballs as *nix (or MacOS) things and why would a Windows developer deal with that?

You should not count on the zip archive provided by GitHub: package maintainers verify downloads against hashes and there [have been issues with GitHub updating their zip library causing the hash to change](https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/).
These days, Windows users have 7zip to extract "exotic" `.tar.xz` archives.
Heck, Windows recently started to use libarchive so these users can open many types of archives from the explorer.

PR #171, opened 2023-10-22
Makefile.am: add src/wavpack.exports to EXTRA_DIST

sezero:
(was missed in c01d06d3f969ae7cd0bb32e95271ca189dce1840)


dbry:
Thanks!

PR #173, opened 2023-11-28
eliminate use of '&' address-of operator on arrays (take/2.)

sezero:
This is an extension of PR #130, where the compiler missed several
cases because of a bug [1]. (These ones were revealed by using the
c++ compiler instead.)

[1] https://github.com/open-watcom/open-watcom-v2/issues/1158


dbry:
Thanks!

PR #187, opened 2024-10-30
cmake: CMAKE_SYSTEM_PROCESSOR is the architecture of the build machine

madebr:
With this patch, I was able to build arm64 Windows executables, for inclusion in SDL3_mixer binary VC releases.

The `wavpackcpu.cmake` script is a copy of SDL's `sdlcpu.cmake`, with "SDL" replaced with `WavPack`.

`CMAKE_SYSTEM_PROCESSOR` is always `AMD64`, even when building wavpack with an arm64 MSVC toolchain.

FYI Microsoft ARM assembler support in CMake uses the `ASM_MARMASM` language.
The MS ARM assembler is not 100% compatible with a GNU ARM assembler.
Luckily, I didn't encounter this issue when building a arm64 wavpack library, but it's something to take into account.
(I didn't test building `src/unpack_armv7.S` for 32-bit arm windows. Windows on 32-bit arm will never happen)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,11 @@ if(CMAKE_ASM_COMPILER)
 endif()
 
 if(MSVC)
-  enable_language(ASM_MASM)
+  if(WavPack_CPU_X86 OR WavPack_CPU_X64)
+    enable_language(ASM_MASM)
+  elseif(WavPack_CPU_ARM32 OR WavPack_CPU_ARM64)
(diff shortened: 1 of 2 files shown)


dbry:
Thanks for this!

At some point I may work on an arm64 version of the ASM code, but not anytime soon. I think the neon instructions might help with stereo decoding speed (like the MMX do), but it's low on my list.

PR #191, opened 2024-11-20
misc cmake stuff

sezero:
(CC: @madebr )


dbry:
Thanks for these, as always!

PR #197, opened 2025-02-06
Add CI

jonaski:
Adds CI which runs `make distcheck`, automatically creates the tarball and Windows binaries. It is triggered on push, pull requests and release, on release it will attach the tarball and the zipped windows binaries to the release. The Windows zip files includes all files that were installed, but this can be adjusted if desired.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: daily
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,217 @@
+name: Build
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  release:
+    types: [published]
+
+jobs:
+
+  tarball:
+    name: Create tarball with distcheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install packages
+        env:
+          DEBIAN_FRONTEND: noninteractive
+        run: sudo apt install -y git tar gcc g++ make autoconf gettext
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          submodules: recursive
+      - name: Add safe git directory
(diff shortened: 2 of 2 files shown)


jonaski:
I see the current binaries are static and also includes the PDF files and changelog. How are the current windows binaries built? Is it with mingw-w64?
If you want, I can skip attaching windows binaries for now and just attach the tar.xz archive.

dbry:
Thanks for doing this!

Yes, I build the official binaries with mingw-w64 because it outperforms the MSVC builds (at least with VS 2019). I think it's good to have MSVC builds going even if I'm not currently using them. I don't mind generating the official binaries and packages by hand because I only do one release a year or so, and I do extra steps like do unix2dos on the text files for the Windows version and replace `config.guess` and `config.sub` with the latest. We could look into all that in the future, but for now I don't think we need Windows packages.

Should I just bring this in and it will start running? I assume you have checked this locally, right? Do you use Docker for that? I'm kind of clueless about this CI stuff. Finally, I guess we'll need to update the README to not show the broken Travis CI anymore.

jonaski:
I use Github Actions on my local fork: https://github.com/jonaski/WavPack/actions/runs/13189185817
If you enable GitHub actions on this repo, and accept to run it for this PR you can see this before you merge too.

dbry:
Hmm. I went to Settings -> Actions -> General and "Allow all actions and reusable workflows" is enabled already. I can't find anywhere to accept running it for this PR. Maybe I need to just pull it in?

jonaski:
Should be a button at the bottom of this PR that you need to accept for actions to run for new contributors. I re-pushed now, do you see it now?

dbry:
no-actions

No, but it says "Continuous integration has not been set up" even though GitHub actions have always been enabled.

jonaski:
I see, maybe it needs to run on the main branch first, but it's OK for me to merge this, I've tested it on my fork so it should be good.

dbry:
Okay, I'll just merge it and we'll see what happens. I saw the results on your fork and it all seemed correct.

dbry:
It seems to be running the tests. Obviously the libwavpack check takes a while. Thanks again!

PR #198, opened 2025-03-27
common_utils.c: Nullify the correct member

j-mattsson:


dbry:
Thanks for catching this!

BTW, did it actually cause an issue, or is it just an obvious mistake that you saw?

j-mattsson:
> Thanks for catching this!
> 
> BTW, did it actually cause an issue, or is it just an obvious mistake that you saw?

No, wasn't causing any issue, just something I noticed.

dbry:
> > Thanks for catching this!
> > BTW, did it actually cause an issue, or is it just an obvious mistake that you saw?
> 
> No, wasn't causing any issue, just something I noticed.

Cool. Thanks again!

PR #205, opened 2025-06-27
Fix autoreconf with gettext 0.25

alyssais:
Since gettext 0.25, the AM_ICONV macro is installed into the m4 directory, so we need to add that autoconf config directory.  We also need to make sure autoreconf recognizes that it needs to run autopoint to install the gettext macros, which it does heuristically by checking whether AM_GNU_GETTEXT_VERSION is used, so add that and the minimum amount of other stuff required to make it happy.  The version of gettext we say we use shouldn't matter, so I just went with the version in ubuntu:latest since that's what CI uses.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Install packages
         env:
           DEBIAN_FRONTEND: noninteractive
-        run: sudo apt install -y git tar gcc g++ make autoconf gettext
+        run: sudo apt install -y git tar gcc g++ make autoconf autopoint gettext
       - name: Checkout
         uses: actions/checkout@v4
         with:
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,3 +1,5 @@
+SUBDIRS =
+
 pkgconfigdir = $(libdir)/pkgconfig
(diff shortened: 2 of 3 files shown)


dbry:
Thanks!

sezero:
@dbry:  Would you be interested in replacing the AM_ICONV mess with an
in-house implemented detection we've been using in SDL:
https://github.com/libsdl-org/SDL/blob/SDL2/configure.ac#L368-L407
(There is a cmake counterpart to it too.)

sezero:
> @sezero If you think this would be an improvement I would certainly be interested. Thanks for the offer!

See #206 

> 
> I have actually from time to time thought of eliminating the `libiconv` dependancy for the command-line programs. It's only used when specifying or displaying metadata tags on the command line on systems that aren't locally UTF-8. I think all modern Linux distros and MacOS are UTF-8 everywhere. I know that some BSD flavors still use local encodings, but if ISO-8859-1 was handled too (which is also required for ID3v2 tags), that would cover just about everyone.
> 
> I don't know if this would be worth it, but libiconv has certainly been a constant issue as long as I can remember! Any perspective on this would be appreciated, especially as someone from Turkey! 😄

I don't have hard feelings on the matter. Maybe make it optional??

PR #206, opened 2025-07-09
revise iconv() detection:

sezero:
- remove iconv.m4 (and by extension gettext) dependency
  from autotools, and remove Iconv find_package() from cmake.
- check libiconv with a linkage test with iconv.h included
- check libc iconv with a linkage test with iconv.h included
  and LIBICONV_PLUG defined (in case libiconv header is in
  include path)
- add new configuration option to prefer iconv from libiconv,
  if available, over the libc version, defaults to disabled:
 ` -DWAVPACK_ENABLE_LIBICONV=ON` for cmake, and `--enable-libiconv`
  for autotools.


dbry:
Tested by building successfully on Raspberry Pi both with Autotools and with cmake. Thanks!

PR #234, opened 2026-06-01
fix signed overflow in non-syncsafe id3v2.3 frame size

aizu-m:
UBSan (clang -fsanitize=shift) on an ID3v2.3 tag imported via `wvtag --import-id3`:

    cli/import_id3.c:181:116: runtime error: left shift of 255 by 24 places
      cannot be represented in type 'int'

`frame_header[4]` promotes to int, so a high size byte makes `frame_size` negative, slips past the `frame_size > tag_size` check, and reaches `memcpy(frame_body, tag_data, frame_size)` with a huge size_t. Cast the top byte like the other 32-bit assemblers in the tree and reject a negative size.

### DIFF
--- a/cli/import_id3.c
+++ b/cli/import_id3.c
@@ -178,14 +178,14 @@ static int ImportID3v2_syncsafe (WavpackContext *wpc, unsigned char *tag_data, i
         if (syncsafe)
             frame_size = frame_header [7] + (frame_header [6] << 7) + (frame_header [5] << 14) + (frame_header [4] << 21);
         else
-            frame_size = frame_header [7] + (frame_header [6] << 8) + (frame_header [5] << 16) + (frame_header [4] << 24);
+            frame_size = frame_header [7] + (frame_header [6] << 8) + (frame_header [5] << 16) + ((uint32_t) frame_header [4] << 24);
 
         if (!frame_size) {
             sprintf (error, "invalid %s tag (empty frame encountered)", tag_type);
             return -1;
         }
 
-        if (frame_size > tag_size) {
+        if (frame_size < 0 || frame_size > tag_size) {
             sprintf (error, "invalid %s tag (truncated)", tag_type);
             return -1;
         }


dbry:
Thanks for the catch, and the PR!

I wonder if a slightly cleaner way of fixing this would be to simply declare `frame_size` as a `uint32_t`. What do you think?

sezero:
>  simply declare `frame_size` as a `uint32_t`. What do you think?

The UB issue is frame_header[n] being promoted to signed int, and one might even add uint32_t casts to not just [4] but to all indices of it when shifting if he is overzealous enough, i.e. you can make it more complex instead of simpler:)

So, I don't think changing frame_size to unsigned would help with the UB at all. On the other hand, keeping frame_size as signed would be good for negative value checks (which I think is a good addition here) without using a signed cast.

dbry:
Yeah, the `uint32_t` cast is needed for the UB trigger, I've got those all over the place (unfortunately).

But the _bug_ here is not checking for negative `frame_size`, which would be redundant if it was unsigned. However,  changing that to unsigned without changing other variables might generate other warnings (I think VS will sometimes warn if you compare a signed with an unsigned).

Thanks to you both, I'll apply the patch as is.

sezero:
>  not checking for negative `frame_size`, which would be redundant if it was unsigned

If you expect sizes > INT32_MAX to be legitimate, then yes

dbry:
> > not checking for negative `frame_size`, which would be redundant if it was unsigned
> 
> If you expect sizes > INT32_MAX to be legitimate, then yes

I'm not sure I'm following here, but the whole ID3 tag cannot be larger than 256MB, so any frame size over that is  definitely _illegitimate_. Another fix for both the UB and the bug might have been:

```
frame_size = frame_header [7] + (frame_header [6] << 8) + (frame_header [5] << 16) + ((frame_header [4] & 0x7f) << 24);
```
Except that would have left a significant error in the tag unflagged.

sezero:
> I'm not sure I'm following here, but the whole ID3 tag cannot be larger than 256MB, so any frame size over that is definitely _illegitimate_.

Yes, exactly: That's why keeping it signed + doing a negative value check is good for rejecting corrupt/malicious files. (But if you have checks elsewhere e.g. against 256MB -- I haven't looked -- then the signedness/negative value check would be unnecessary I guess.)

dbry:
I'm sure we're saying the same thing. But yes, the way the tag size is calculated at this line prevents anything over 256MB, and everything is later tested against that:

https://github.com/dbry/WavPack/blob/01b70138e9a4ad4a21fb8f783f3a99d8fc56241c/cli/import_id3.c#L139

And if the value is unsigned, then checking against a maximum (which we do) is all that's required. An unsigned variable can't be negative; malicious input would just result in something huge and trip the max check.

But again, if you mix signed and unsigned variables you can end up generating a lot of warnings, even if you know what you're doing.

sezero:
OK, agreed.

PR #235, opened 2026-06-02
fix signed shift overflow in WavpackOpenRawDecoder header parsing

aizu-m:
UBSan from clang -fsanitize=shift, driving WavpackOpenRawDecoder with Matroska-style raw blocks (no "wvpk" header) whose reconstructed header bytes have the high bit set:

    open_raw.c:163:41: runtime error: left shift of 128 by 24 places
      cannot be represented in type 'int'

The bytes promote to int before the `<< 24`, so any top byte >= 0x80 overflows the sign bit. Cast to uint32_t for the six block_samples/flags/crc/block_size fields, as WavpackLittleEndianToNative already does.


dbry:
Thanks!

PR #239, opened 2026-06-07
cast parameters of ctype.h calls to unsigned char.

sezero:
Janitorial patch -- hope that I didn't miss an instance.


dbry:
Thanks!

PR #242, opened 2026-06-13
cli/wvunpack.c: bound dump_file_item appends to fix -f stack overflow

aizu-m:
Built the cli tools with ASan and ran `-f` over a crafted file:

    ==ERROR: AddressSanitizer: stack-buffer-overflow
    WRITE of size 2 ...
        #1 dump_file_info wvunpack.c:3081
        #2 unpack_file wvunpack.c:1796
        #3 main wvunpack.c:948
      [32, 112) 'str' (line 3073) <== Memory access overflows this variable

`-f` builds a ';'-separated info line for the nine WavpackGet* fields in an 80-byte `str`. dump_file_item appends each field with `sprintf (str + strlen (str), ...)` and the loop adds a ';' with strcat, none of it bounded. The widths come straight from the .wv header. The MD5 string alone is 32 bytes (present whenever the file carries an ID_MD5_CHECKSUM block), the native sample rate and total-sample count are printed in full, and the channel mask and mode are hex.

An ordinary multichannel file with an MD5 already reaches ~72 bytes. An 18-channel float stream with a ten-digit sample rate and ~1e6 samples tips it past 80 and overruns the stack. Reachable with one command, `wvunpack -f file.wv`.

Pass the buffer size into dump_file_item and bound every append with snprintf, matching the sprintf to snprintf buffer-length fix already in xmms/src/libwavpack.c. After the patch the same file prints a truncated line and runs clean.

### DIFF
--- a/cli/wvunpack.c
+++ b/cli/wvunpack.c
@@ -3066,7 +3066,7 @@ static void dump_summary (WavpackContext *wpc, char *name, FILE *dst)
 // 9. encoding mode (in hex because it's a bitfield, always prefixed with "0x")
 // 10. filename (if available)
 
-static void dump_file_item (WavpackContext *wpc, char *str, int item_id);
+static void dump_file_item (WavpackContext *wpc, char *str, int size, int item_id);
 
 static void dump_file_info (WavpackContext *wpc, char *name, FILE *dst, int parameter)
 {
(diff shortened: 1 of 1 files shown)


dbry:
Thanks for this, good catch! I created a file myself that exceeded 80 characters and verified the fix.

I don't really like the fix though because it adds a lot of complexity to the code when an easier fix would be to simply add 10 or 20 characters to the string size (which is what I would have done). There's nothing special about 80 characters, and in fact I think truncating the string there and providing incorrect information is almost as bad as crashing.

That said, one could argue that in the future someone might add another item and not notice the size issue, and everyone's allergic to `sprintf()`, so I'll bring in your change and then I'll do another change to up the string size to 100 so it doesn't get truncated.

Thanks again!

PR #243, opened 2026-06-13
issue #232: fix strcpy buffer overflow

Mr-Z-2697:
other similar codes are found and fixed as well.

fixes #232

### DIFF
--- a/cli/utils.c
+++ b/cli/utils.c
@@ -482,7 +482,7 @@ void do_pause_mode (void)
 
 void error_line (char *error, ...)
 {
-    char error_msg [512];
+    char error_msg [2048];
     va_list argptr;
 
     error_msg [0] = '\r';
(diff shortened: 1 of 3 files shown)


Mr-Z-2697:
`+ PATH_MAX` is used, silimar to other `malloc`'s related to file names, but it's probably superfluous, since the string is only extended by 1 character in later step.

Mr-Z-2697:
The encoding summary uses `error_line`, it's buffer is 512 bytes and can get truncated when file name is longer.

dbry:
Thanks for this! However, there's an issue. `strlen()` does _not_ include the terminating NULL, and we're adding a character to the string, so the _minimum_ you need to add to the strlen() result is 2. I usually add 10 just so I don't have to think about it (you'll find that everywhere).

It seems like running with the address sanitizer would have caught that every time. Did you try that?

Anyway, if you force-push a fix with those additions changed to 10, I'll pull this in.

Mr-Z-2697:
Thank you, fixed.

PR #244, opened 2026-06-15
validate legacy v3 header shift field in open_file3

aizu-m:
the legacy version 3 decoder copies the header 'shift' field straight into unpack_samples3 and uses it as a bit-shift count on 32-bit samples, so a crafted v3 file with shift >= 32 (or negative) hits an undefined shift while setting up min_value/max_value at unpack3.c:334; reject an out-of-range shift in open_file3 alongside the other header-field checks (regressor added under fuzzing/regression/).


dbry:
Thanks!

PR #246, opened 2026-06-17
src/entropy_utils.c: fix UB (offset to NULL pointer) in read_hybrid_profile

aizu-m:
Ran the regression corpus through the undefined-behaviour sanitiser (the same `-fsanitize=undefined -fno-sanitize=signed-integer-overflow -fno-sanitize-recover` flags `check-undefined` uses):

    src/entropy_utils.c:154:17: runtime error: applying non-zero offset 2 to null pointer
        #0 read_hybrid_profile entropy_utils.c:154
        #1 process_metadata
        #2 WavpackUnpackSamples

`read_hybrid_profile` reads `byteptr` from `wpmd->data` and forms `endptr = byteptr + wpmd->byte_length`. A WV_HYBRID_PROFILE block with a zero byte_length leaves `wpmd->data` NULL, since `read_metadata_buff` returns NULL data for empty blocks. The bounds tests then evaluate `NULL + 2` before anything checks the pointer, which is the offset-on-null UB above. Trigger is a crafted `.wv` reaching `process_metadata` during block parsing.

Same shape as `read_decorr_samples`, which you guarded in fd1531c. This is the sibling reader that was missed, so I matched that fix. Added a 434-byte regression file that aborts `check-undefined` before the patch and runs clean after.


dbry:
Looks good, thanks!

PR #247, opened 2026-06-17
reject pre-3.0 files with non-16-bit samples in open_file3

aizu-m:
UBSan, `--enable-legacy` build:

    src/unpack3.c:334:53: runtime error: shift exponent -8 is negative

Was going over the pre-3.0 decode path. open_file3 derives the sample shift
for version 1/2 files as `16 - BitsPerSample`, but the WAV header check only
bounds BitsPerSample to [16,24]. Versions before 3.0 only stored 16-bit
samples (BYTES_3 and OVER_20 are version 3 flags, and they get dropped when
the flags are recomputed for these files), so anything above 16 leaves shift
negative.

unpack_samples3 uses shift as a shift count straight away (`min_value =
-32768 >> shift`, then `sample << shift`), so a crafted mono version-2 file
declaring BitsPerSample 24 aborts on the first WavpackUnpackSamples call,
before any sample is even read. Only reachable on legacy builds, which the
fuzzer does not cover.

Reject pre-3.0 files whose BitsPerSample is not 16. Added a regression file
under fuzzing/regression for the legacy sanitiser regressors; it aborts the
check-undefined regressor before the patch and decodes (is rejected) cleanly
after.


dbry:
Thanks for this catch! Actually the UB was interesting, but it led me to another discovery. Version 2.0 files, and some version 3.X files can be from WAV files indicating 9 - 15 bits, but the exsiting code was rejecting those. That's why when we calculated the shift we used BitsPerSample from the WAV header. If only 16 bits was allowed, then that line wouldn't have made sense.

Anyway, I have commited a fix (2115fa930e8c95a5ee010df4fdef5b777a40b4ce) that allows files from 9 - 15 bits and fixes the UB you found by checking for BitsPerSample > 16.

Thanks!

aizu-m:
Good point, I read the [16,24] header bound as the whole story and didn't realise 9-15 bit legacy files were legitimate, so I drew the line too tight. `> 16` is the right cut since that's the only range that turns the derived shift negative, and it keeps the BitsPerSample line in the shift calc meaningful. Thanks for sorting it.

PR #249, opened 2026-06-19
cast sample output operands to uint32 in legacy unpack_samples3

aizu-m:
UBSan, --enable-legacy build, run over the fuzz regression corpus:

    src/unpack3.c:412:41: runtime error: left shift of negative value -1

unpack_samples3 stores each decoded sample with `value << shift`, where shift is the version-3 header field and the operand is a signed sample that is routinely negative for real audio. Left shift of a negative value is undefined behaviour, and it trips even when shift is 0. The min/max clamp operands just above already cast through uint32, but the sample stores did not, so the regressor still aborted here.

Two sites tripped first (412 in the fast mono path, 830 in the new-high path). The remaining sample-output shifts in the function share the defect, so I cast all of them, matching the existing min_shifted form a few lines up.

This path is not built by the OSS-Fuzz target. The added regression file aborts the check-undefined regressor before the change and decodes clean after.


dbry:
I'm not sure I understand the purpose of the regression file. _Any_ use of this code to decode legacy audio triggers UB, both left-shift of negative and signed integer overflow (in the crc calculation). I am aware of these and haven't bothered because it's legacy code that _should not be built into any library_ and is only used in rare cases to recover archived audio. And these UB instances currently pose no risk and work fine on every modern processor.

That said, I can go ahead and fix both issues because it can be done without performance hit or risk of breakage.

dbry:
Fixed in 0481f63827f7369cab13893e188d6d868f1c7179

aizu-m:
Fix looks complete. Every sample store is cast now, and moving crc to uint32 kills the checksum overflow too, which my patch didn't touch.

On the regression file: the only reason for it was that the legacy decode path gets run under UBSan when the regression corpus is exercised, so a committed input would keep that abort from creeping back in. But you're right that the whole path is UB by design, and you've fixed it at source, so the file is moot. Thanks for sorting both.

PR #260, opened 2026-07-22
Reject invalid raw decoder buffer lengths

acts-1631:
WavpackOpenRawDecoder compared the raw signature before checking the main buffer length. A truncated raw frame could read beyond the caller-provided buffer and crash an application decoding untrusted media.

Validate main and correction buffer arguments before parsing. This rejects missing or too-short main buffers, negative correction lengths, and nonzero correction lengths without a correction buffer.

Verified with ASan/UBSan, the truncated-frame reproducer, invalid correction-buffer inputs, targeted raw-frame fuzzing, and the existing regression corpus.

### DIFF
--- a/src/open_raw.c
+++ b/src/open_raw.c
@@ -137,6 +137,11 @@ WavpackContext *WavpackOpenRawDecoder (
 {
     WavpackRawContext *raw_wv = NULL, *raw_wvc = NULL;
 
+    if (!main_data || main_size < 4 || corr_size < 0 || (corr_size && !corr_data)) {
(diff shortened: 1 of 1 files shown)


dbry:
Thanks for the report and patch!

I would not normally consider this an issue because any reasonable user of this library function would catch such egregious errors before calling, and would certainly never pass in NULL pointers with valid counts. _Maybe_ some code would be so braindead as to pass in a buffer less than 4 bytes, but even that seems really dubious.

That said, this doesn't hurt anything, so I'll pull it in.

